Issue #7, opened 2025-05-17
Add docker


mosonyi:
Create a Dockerfile to enable Docker container to run the application.

PR #8, opened 2025-05-17
Added docker env #7

mosonyi:


clangenb (approved):
Nice it works. The image is kind of big though, but it does not seem that there is anything unnecessary.

Issue #9, opened 2025-05-17
Add gitlab ci


mosonyi:
At least:
 
* lint
* build

PR #10, opened 2025-05-17
Added ci build #9

mosonyi:


mosonyi:
@clangenb , please fix the errors so we can merge them.

Issue #62, opened 2025-10-31
Test-Swarm setup for testing


clangenb:
So far we have only been reading info from a swarm so we could just check a running set up.

However, if we start to implement commands we should have a test-setup, which is safe to play around with.

clangenb:
Closed by #64

PR #64, opened 2025-10-31
Added test environment #62

mosonyi:
Added a small test env with README.


clangenb:
Cool. I can reproduce it. I can see the service whoami on one node, but the stacks seem to be empty. Can you expand this to also create a stack?

PR #66, opened 2025-11-01
Fix Integration tests

clangenb:
This gives us a proper integration test setup:

1. Start a dind swarm in the CI
2. deploy a stack
3. Run the first integration test that actually calls the docker API and verifies that the stack called `demo` is deployed.
4. The bash script also works locally.

I would like to merge this functional version first before I introduce more sophisticated tests if you agree with my docker-compose changes @mosonyi.


clangenb:
Merging this, as I want to continue elsewhere.

PR #67, opened 2025-11-03
Add Docker methods for restarting a service safely (scale down, and then scale up)

clangenb:
Adds scaling and restarting logic to the docker library, along with some integration tests (no UI integration yet).

We need to restart it safely via the scale down and scale up approach to be absolutely sure that all services are scaled stopped before starting them again to be absolutely sure to prevent double signing.

### DIFF
--- a/test-setup/test-stack.yml
+++ b/test-setup/test-stack.yml
@@ -1,5 +1,3 @@
-version: "3.8"
-
 services:
   whoami:
     image: traefik/whoami:v1.10


mosonyi:
@clangenb, I made a mistake when I defined the original issue. Here's the summary:
For restarting, you should use the first one.

| Task | Command | Notes |
|------|----------|-------|
| Restart service | `docker service update --force <name>` | Safest, rolling restart |
| Update image | `docker service update --image newimage:tag <name>` | Deploys new image |
| Restart all replicas instantly | `docker service scale <name>=0 && docker service scale <name>=N` | Causes downtime |
| Customize update behavior | `--update-parallelism`, `--update-delay`, `--update-failure-action` | Fine-tunes rollout |

PR #129, opened 2025-12-05
Add docker context handling features

mosonyi:
type `:ctx` or `:context` to go to the context 
All the options are available
* import
* create (even with tls)
* export
* update
* delete
* inspect

@clangenb please note: docker context management is a CLI-only feature and is not available in the Docker Go SDK

### DIFF
--- a/app/update.go
+++ b/app/update.go
@@ -119,8 +119,8 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			cmd,
 		)
 
-	case loadingview.NavigateToContextsMsg:
-		// Navigate to contexts view from error screen
+	case loadingview.ErrorDismissedMsg:
+		// Navigate to contexts view from loading error screen
 		cmd := m.replaceView(contextsview.ViewName, nil)
 		return m, tea.Batch(
 			cmd,
--- a/views/loading/loading.go
+++ b/views/loading/loading.go
@@ -13,8 +13,8 @@ import (
 
 const ViewName = "loading"
 
-// NavigateToContextsMsg is sent when user presses Enter on error
-type NavigateToContextsMsg struct{}
+// ErrorDismissedMsg is sent when user presses Enter on an error screen
+type ErrorDismissedMsg struct{}
 
 type Model struct {
 	width, height int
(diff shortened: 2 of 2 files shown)


commands/command/contexts.go
```diff
@@ -0,0 +1,45 @@
...
+	// Register aliases
+	registry.Register(aliasCommand{name: "context", target: contextsCmd})
+	registry.Register(aliasCommand{name: "ctx", target: contextsCmd})
+}
+
+// aliasCommand is a simple wrapper to provide aliases for commands
+type aliasCommand struct {
+	name   string
+	target registry.Command
+}
```
clangenb: aliasCommand was a Copilot example I suppose. This should be `ContextsCmd`

docker/context.go
```diff
@@ -0,0 +1,444 @@
...
+	cmd := exec.Command("docker", "context", "use", contextName)
+	// Don't output to stdout/stderr to keep UI clean
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to switch context to %s: %w", contextName, err)
+	}
+	return nil
+}
+
+// ValidateContext checks if a context switch would succeed by attempting to connect
+func ValidateContext(contextName string) error {
```
clangenb: I think this method is unused?

docker/context.go
```diff
@@ -0,0 +1,444 @@
...
+	Name        string
+	Current     bool
+	Description string
+	DockerHost  string
+	TLS         bool
+	Error       string
+}
+
+// ListContexts returns all available Docker contexts using docker CLI
+func ListContexts() ([]ContextInfo, error) {
```
clangenb: Can you please ask if this function is idiomatic? This looks much more complex than necessary. Please always ask for generated code to be: idiomatic, according to best practice and strongly typed.

docker/context.go
clangenb: In general, this file looks good file. But maybe you can ask our friend to make the entire file more idiomatic. I have added some specific hints.

views/contexts/view.go
```diff
@@ -0,0 +1,728 @@
...
+			currentLine = word
+		}
+	}
+
+	if currentLine != "" {
+		lines = append(lines, currentLine)
+	}
+
+	return lines
+}
```
clangenb: All those dialogs are very generic. I suggest moving them to the ui/components package, but you can skip this now. Just please do not duplicate them when you need them somewhere else.

mosonyi:
> I did test the contexts view, and I noticed that the cursor automatically switches back to the current context after a few seconds.
Fixed

commands/command/contexts.go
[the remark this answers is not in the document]
mosonyi: Moved, done

docker/context.go
clangenb: I actually noticed that the docker's CLI version allows switching to unusable contexts, and it was a pain.

clangenb (approved):
Thanks for the fixes, looks much better now. There are some follow-ups, but I think we can merge this now.

PR #146, opened 2025-12-13
Enhance frame view

mosonyi:
How to test:
* check all the views if 100% of the width and height are used
* on all screen type ':' to see if the command frame breaks anything
* Resize the windows
* On all views try to manipulate the docker from another window and see if the change  appears
  for example:
  * on stack view `docker stack rm demo`
  * on service view `docker service update --replicas 5 <service_name>`
  * on node view `docker node update --label-add foo=asd <node_id>`

### DIFF
--- a/views/nodes/model.go
+++ b/views/nodes/model.go
@@ -1,8 +1,6 @@
 package nodesview
 
 import (
-	"encoding/json"
-	"os"
 	"strings"
 	"swarmcli/core/primitives/hash"
 	"swarmcli/docker"
@@ -99,30 +97,7 @@ func CheckNodesCmd(lastHash uint64) tea.Cmd {
 		l().Infof("CheckNodesCmd: lastHash=%s, newHash=%s, nodeCount=%d",
 			hash.Fmt(lastHash), hash.Fmt(newHash), len(entries))
 
-		// Write debug dump to /tmp for quick inspection during troubleshooting
-		go func() {
-			type dump struct {
-				Last    string              `json:"last"`
-				New     string              `json:"new"`
-				Count   int                 `json:"count"`
-				Entries []map[string]string `json:"entries"`
-			}
-			d := dump{Last: hash.Fmt(lastHash), New: hash.Fmt(newHash), Count: len(entries), Entries: []map[string]string{}}
-			for i, e := range entries {
-				if i >= 10 {
-					break
-				}
-				labels := ""
-				if e.Labels != nil {
-					// Marshal labels into JSON string for readability
-					lb, _ := json.Marshal(e.Labels)
-					labels = string(lb)
-				}
-				d.Entries = append(d.Entries, map[string]string{"ID": e.ID, "Hostname": e.Hostname, "Labels": labels})
-			}
-			b, _ := json.MarshalIndent(d, "", "  ")
-			_ = os.WriteFile("/tmp/swarmcli_nodes_check.json", b, 0644)
-		}()
+		l().Debugf("CheckNodesCmd: Node entries: %+v", entries)
 
 		// Only return update message if something changed
 		if newHash != lastHash {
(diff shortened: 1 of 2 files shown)


docker/events.go
clangenb: Very important feature - I like it.

However, the docker library structure is becoming a bit convoluted. We need to introduce proper abstraction layers: https://github.com/mosonyi/swarmcli/issues/148

views/contexts/model.go
```diff
@@ -179,6 +234,12 @@ func (m *Model) MoveCursor(delta int) {
 	if m.cursor >= len(m.contexts) {
 		m.cursor = len(m.contexts) - 1
 	}
+	// Mirror to internal FilterableList cursor and ensure visible
+	if m.List.Cursor != m.cursor {
+		m.List.Cursor = m.cursor
+		// ApplyFilter will keep the cursor in-bounds and update viewport offset
+		m.List.ApplyFilter()
+	}
```
clangenb: Why do we even have separate cursors? This makes everything so much more complicated, and I cannot see the benefit: https://github.com/mosonyi/swarmcli/issues/149

views/contexts/model.go
```diff
@@ -194,6 +255,14 @@ func (m *Model) SetLoading(loading bool) {
...
 	m.loading = loading
+	// Write debug state when loading changes
+	go m.debugWriteLoadingState()
+}
+
+// debug write of loading state for diagnostics
+func (m *Model) debugWriteLoadingState() {
+	// best-effort write; don't fail on error
+	_ = os.WriteFile("/tmp/swarmcli_contexts_state_debug.json", []byte(fmt.Sprintf("{\"loading\":%v,\"count\":%d}\n", m.loading, len(m.contexts))), 0644)
 }
```
clangenb: Why do we not just log it?

views/services/update.go
```diff
@@ -180,100 +179,120 @@ func (m *Model) SetContent(msg Msg) {
...
+			colWidths[i] = width - starts[i]
+		} else {
+			colWidths[i] = starts[i+1] - starts[i]
 		}
-		if len(e.StackName) > maxStack {
-			maxStack = len(e.StackName)
+		if colWidths[i] < 1 {
+			colWidths[i] = 1
 		}
 	}
```
clangenb: Third time that I see this kind of code now. The header should be integrated into the FilteredList itself: https://github.com/mosonyi/swarmcli/issues/150

clangenb (changes requested):
Feature wise, I think it looks awesome. The UI looks great and is snappy.

I noticed one little bug: the height of the inspect view is off.

There is one change that I would like to request: Please remove the debug dumps. It is very confusing to investigate all those different files. I prefer to just have them all in the log file, then we can also better see when the logs have occurred relative to other operations during the runtime.

Additionally, the UI code is hard to maintain like this. It introduces manual UI fixing at many places in order to fix the layout. If we change something somewhere, we need to fix it at many places. It is better to introduce proper (centralized) abstractions that handle the layout, i.e. instead of fixing the view methods of contexts/nodes/stacks manually with respect to the padding - we should introduce a padding parameter in the ListView and let it handle the padding: https://github.com/mosonyi/swarmcli/issues/151.

Then the flow would look like this:
1. We introduce a global padding constant that the root viewport knows.
2. It passes the padding parameters to its current views
3. They use this parameter to handle the padding of their content, or if they use the FilteredList, they just forward this value to the list.

However, this would be a major refactor, and I think it should not be in this PR, which is already quite big. I just put this example here, as I think that the PR is a very good example on how we would benefit from this architectural change, and us the opportunity to explain the philosophy.

views/contexts/model.go
mosonyi: See https://github.com/mosonyi/swarmcli/pull/146/commits/2e0313785490b5c22e1a443f3675dfa9e542fac0

PR #158, opened 2026-01-02
Several UI fixes

mosonyi:
* Ensure that in all views the frame width & height is 100%.
* From the stack view (the default main view), you cannot exit the program with Esc.
* If you filter any list, the Esc should clear the filter only.
* If you filter on anything, the control characters overwrote the filter. For example at service view, you coouldn't search for string contains 'l' or 'q'

### DIFF
--- a/ui/framecalc.go
+++ b/ui/framecalc.go
@@ -32,7 +32,6 @@ func ComputeFrameDimensions(viewportWidth, viewportHeight, fallbackWidth, fallba
 	if frameWidth <= 0 {
 		frameWidth = 80
 	}
-	frameWidth += 4
 	frameWidth += horizontalPadding
 
 	frameHeight := viewportHeight


ui/framecalc.go
```diff
@@ -0,0 +1,59 @@
...
+// lines as frameHeight - borders - header - footer (never negative).
+func ComputeFrameDimensions(viewportWidth, viewportHeight, fallbackWidth, fallbackHeight int, header, footer string) FrameSpec {
+	frameWidth := viewportWidth
+	if frameWidth <= 0 {
+		frameWidth = fallbackWidth
+	}
+	if frameWidth <= 0 {
+		frameWidth = 80
+	}
+	frameWidth += 4
```
clangenb: let's define a `const horizontalPadding = 4`.

And then do

```suggestion
	frameWidth += horizontalPadding
```

ui/framecalc.go
```diff
@@ -0,0 +1,59 @@
...
+	headerLines := 0
+	if header != "" {
+		headerLines = len(strings.Split(header, "\n"))
+	}
+	footerLines := 0
+	if footer != "" {
+		footerLines = len(strings.Split(footer, "\n"))
+	}
+
+	desiredContentLines := frameHeight - 2 - headerLines - footerLines
```
clangenb: What's the two, vertical padding?

ui/ui.go
```diff
@@ -167,6 +167,27 @@ func RenderFramedBoxHeight(title, header, content, footer string, width, frameHe
 	return RenderFramedBox(title, header, paddedContent, footer, width)
 }
 
+// TrimContentToLines returns content limited to exactly `lines` rows,
+// padding with empty lines when shorter. Useful when framing viewport
+// content to a fixed height.
+func TrimContentToLines(content string, lines int) string {
```
clangenb: Let's call it `TrimOrPadContentToLines` fits the purpose better.

views/configs/model.go
```diff
@@ -20,6 +20,7 @@ type Model struct {
 	configsList  filterlist.FilterableList[configItem]
 	width        int
 	height       int
+	firstResize  bool   // tracks if we've received the first window size
```
clangenb: I think it would be more logical to invert the logic here, i.e.:

`didFirstResize bool` and initialize it with false. This is much more intuitional.

views/configs/update.go
```diff
@@ -56,6 +56,13 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 		// msg.Height is already adjusted by the app to account for the
 		// systeminfo header; avoid subtracting extra lines here.
 		m.configsList.Viewport.Height = msg.Height
+		// On first resize, reset YOffset to 0; on subsequent resizes, only reset if cursor is at top
+		if m.firstResize {
+			m.configsList.Viewport.YOffset = 0
+			m.firstResize = false
+		} else if m.configsList.Cursor == 0 {
+			m.configsList.Viewport.YOffset = 0
+		}
```
clangenb: The logic looks good, but what problem is this solving?

views/configs/model.go
clangenb: It is ok. We can leave it at that. I believe that we will fix this issue better in the future anyhow. :)

clangenb (commented):
Nice, I like the PR a lot. There are only a few tiny details to fix! :)

ui/framecalc.go
mosonyi: See https://github.com/mosonyi/swarmcli/pull/158/commits/7a70dc59e30c7ba66bab8726aa6e1c0b35c5074d

ui/framecalc.go
```diff
@@ -0,0 +1,65 @@
...
+// lines as frameHeight - vertical padding - header - footer (never negative).
+func ComputeFrameDimensions(viewportWidth, viewportHeight, fallbackWidth, fallbackHeight int, header, footer string) FrameSpec {
+	frameWidth := viewportWidth
+	if frameWidth <= 0 {
+		frameWidth = fallbackWidth
+	}
+	if frameWidth <= 0 {
+		frameWidth = 80
+	}
+	frameWidth += 4
```
clangenb: You added my suggestion, but did not remove the previous version ;)

Issue #159, opened 2026-01-02
Restart service


mosonyi:
The restart service should not wait until the service restarts.

PR #160, opened 2026-01-08
Restart service fix

mosonyi:
Fixing #159 

Fixes
- When I press 'r' at the pop-up, the esc was handled on the top level.
- If the service restart failed, there was no error message
- After the restart, the pop-up hasn't closed.

### DIFF
--- a/docker/snapshot.go
+++ b/docker/snapshot.go
@@ -45,7 +45,7 @@ var (
 )
 
 // cacheTTL controls how long we reuse the snapshot before refreshing.
-const cacheTTL = time.Minute
+const cacheTTL = 10 * time.Second
 
 // GetSnapshot returns the cached snapshot if it's still valid.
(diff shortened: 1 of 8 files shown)


docker/snapshot.go
```diff
@@ -45,7 +45,7 @@ var (
 )
 
 // cacheTTL controls how long we reuse the snapshot before refreshing.
-const cacheTTL = time.Minute
+const cacheTTL = 10 * time.Second
```
clangenb: This will cause a lot of refreshes, are you sure that you want this value?

clangenb (approved):
I have one question, but it looks good to me otherwise 👍

docker/snapshot.go
mosonyi: Yes, it is needed. If I restart the service, I would like to see in almost real time when it is up again.

Issue #44, opened 2025-10-27
Command: secret


mosonyi:
Default: `docker secret ls`

Controlls:
'c' create new: `docker secret create [OPTIONS] SECRET [file|-]` (see docker secret create --help)
'i' inspect: `docker secret inspect <ID>`
'ctrl+d' rm: `docker secret rm <ID>` (Note: if the secret is used, it cannot be deleted)

PR #169, opened 2026-01-12
Added all secret options.

mosonyi:
#44 

How to test:
type ":secret"

By default docker swarm doesn't allow to reveal secrets like kubernetes, but with a small trick I made it available.
If the docker-proxy-rbac will be in placed we can restrict these kind of access.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -1,12 +1,17 @@
 package docker
 
 import (
+	"bytes"
 	"context"
+	"encoding/binary"
 	"fmt"
+	"io"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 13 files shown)


docker/service.go
```diff
@@ -509,3 +514,183 @@ func getServiceStatus(svc swarm.Service) string {
...
+				Condition: swarm.RestartPolicyConditionNone,
+			},
+		},
+		Mode: swarm.ServiceMode{
+			Replicated: &swarm.ReplicatedService{
+				Replicas: func() *uint64 { r := uint64(1); return &r }(),
+			},
+		},
+	}
+}
```
clangenb: Cool, trick 😎

clangenb (approved):
Cool. Looks good to me!

PR #197, opened 2026-02-02
Fix ps fatal error and add enchacements

### LINKED ISSUE #196: [Fatal error] By stack ps if "no suitable node found"
```docker service ps <service_name>```
or
```docker stack ps <stack_name>```

can show us if there's an error -> ex: ```Running         Pending 3 days ago      "no suitable node (scheduling …"```

We need to reveal if there's an error.

Currently, what we could see is "1/1             updating       replicated"

If I press 'p' in the service view to check it, the program crashes.

Fixes:
1. At stack view, add a ps, but in the background search for errors and show by default
2. Same on service level.

### PR DESCRIPTION
**Pull request by mosonyi**
Fixes #196 

* Fixes fatal error
* Show clearly the error if there's an error deploying on the service level

### DIFF
--- a/docker/snapshot.go
+++ b/docker/snapshot.go
@@ -15,15 +15,16 @@ import (
 )
 
 type NodeEntry struct {
-	ID           string
-	Version      string
-	Hostname     string
-	Role         string
-	State        string
-	Availability string
-	Manager      bool
-	Addr         string
-	Labels       map[string]string
+	ID            string
+	Version       string
+	Hostname      string
+	Role          string
+	State         string
+	Availability  string
+	Manager       bool
+	Addr          string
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/view.go [hunk 1]
```diff
@@ -27,10 +27,16 @@ func (m *Model) View() string {
 		globalHelp = []helpbar.HelpEntry{}
 	}
 
+	// Check if current view has errors (for stacks view)
```
clangenb: Wouldn't it be more future-proof if we add the `HasErrors` to the generic view interface?

And then we can simply call `HasErrors` here. For views where we do not have a specific error handling defined yet, we will just return `false.

What do you think?

clangenb (approved):
Cool, looks good to me now!

PR #201, opened 2026-02-03
Fix search in logs function 

### LINKED ISSUE #200: [logsView] can't use `s` in search (it still toggles log following)

mosonyi:
Fixes #200


clangenb (commented):
Looks good in general, but why do we need this here?

PR #204, opened 2026-02-10
Change readme and docs

mosonyi:


CONTRIBUTING.md
clangenb: We already have a Contributing.md. If you think this is better delete the old one.

clangenb (commented):
Some minor comments, but looks good otherwise.

PR #240, opened 2026-02-18
Inline command aliases in help view

### LINKED ISSUE #191: [help] inline command alias instead of repeating entries
I think it is better to add extra columns for aliases, i.e., instead of three entries like:

│ :ctx            |  List and switch Docker contexts         │
│ :contexts   |      List and switch Docker contexts     │
│ :context     |     List and switch Docker contexts      |

we should have:

│ :ctx            |  List and switch Docker contexts         │ aliases: contexts, context

eldara-cruncher:
## Summary

- Aliases (e.g. `ctx`, `context`) no longer appear as separate rows in `:help`
- They are grouped with their primary command and rendered as a three-column table:
  ```
  COMMAND          DESCRIPTION                          ALIASES
  :contexts        List and switch Docker contexts       ctx, context
  :help            Show all available commands
  :stacks          List stacks
  ```
- Column widths computed dynamically from data; 3-space gap between columns
- Header styled bold blue (lipgloss color 33, matching categorized help keys)
- Adds `Aliaser` interface to registry so alias commands can self-identify

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./registry/ ./commands/command/ ./views/help/` — all pass
- [x] Verify `:help` output renders aligned table with blue header

Closes #191

### DIFF
--- a/commands/command/alias.go
+++ b/commands/command/alias.go
@@ -21,3 +21,6 @@ func (a aliasCommand) Description() string { return a.target.Description() }
 func (a aliasCommand) Execute(ctx any, args args.Args) tea.Cmd {
 	return a.target.Execute(ctx, args)
 }
+
+// AliasOf implements registry.Aliaser.
+func (a aliasCommand) AliasOf() string { return a.target.Name() }
--- a/commands/command/alias_test.go
+++ b/commands/command/alias_test.go
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package command
+
+import (
+	"testing"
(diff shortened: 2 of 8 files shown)


clangenb:
It works, but the layout does not look too nice.

image


I think it would be better if we can have a table with the headers: | Available Commands | Descriptions | aliases |, and the columns are nicely aligned to a meaningful width.

clangenb (commented):
LGTM. @mosonyi please also have a look to see if you approve.

clangenb (approved):

mosonyi:
I would add a space after the frame |<space><content>

PR #251, opened 2026-02-21
fix: place service labels under deploy.labels in stack YAML

### LINKED ISSUE #248: Edit stack
When I edit a stack yaml:
The labels are not below the deploy section where they're belong.

eldara-cruncher:
## Summary
- Service labels (`ServiceSpec.Labels`) are now correctly placed under `deploy.labels` instead of at the service level in reconstructed stack YAML
- Container labels (`ContainerSpec.Labels`) remain at the service level where they belong
- Docker-internal labels (`com.docker.stack.*`) are filtered from both locations

Closes #248

## Test plan
- [x] Unit tests for `filterLabels` helper (5 cases: Docker keys removed, user keys kept, all-filtered returns nil, empty map, nil input)
- [x] YAML structure tests verifying deploy labels, service labels, and both together produce correct output
- [x] Full test suite passes

### DIFF
--- a/docker/stack_to_compose.go
+++ b/docker/stack_to_compose.go
@@ -261,11 +261,8 @@ func ReconstructStackCompose(stackName string) (string, error) {
 		key := sanitizeServiceName(stackName, si.Spec.Name)
 		cs := ComposeService{}
 
-		// Collect labels from service and container specs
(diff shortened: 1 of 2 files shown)


mosonyi (approved):
Tested, looks fine.

PR #258, opened 2026-02-27
Allow import .dockercontext

### LINKED ISSUE #257: Context import .dockercontext files
Allow to import not only tar files but .dockercontext .tar .tgz .tar.gz

mosonyi:
Fixes #257

### DIFF
--- a/docker/context.go
+++ b/docker/context.go
@@ -15,6 +15,10 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


clangenb (commented):
I agree with the review, please fix the duplicate as noted in 1. The statements 2,3 are actually compliments ;)

PR #266, opened 2026-03-05
Fix tty: true logs 

### LINKED ISSUE #265: Logs view closes immediately for TTY Swarm services
Logs view closes immediately for TTY Swarm services ([--- stream closed ---] even though [docker service logs --raw] returns data

### Summary
For services created with tty: true, the logs screen in [swarmcli] shows only [--- stream closed ---] and no log lines.
The same service returns logs correctly from CLI with [docker service logs --raw --follow].

### Environment

OS: Linux (Debian dev container)
App: [swarmcli] (current workspace version)
has tty: true
Steps to reproduce

Deploy a Swarm service with tty: true.
Open that service in [swarmcli] logs view.
Observe output.
Compare with CLI:
[docker service logs --raw --follow --details --tail all <servicename>]

### Actual behavior

[swarmcli] logs view shows only [--- stream closed ---].
No log lines are rendered in the TUI.
Expected behavior

[swarmcli] should stream logs for TTY services the same way as CLI raw mode, without immediate stream closure.
Observed diagnostic clue

Docker returns/relates to: tty service logs only supported with --raw.

SDK stream path uses demux [stdcopy] assumptions that can fail on raw TTY output; stream then ends and UI emits [--- stream closed ---] instead of preserving/continuing raw stream output.

mosonyi:
Fixes #265

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -9,6 +9,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"os/exec"
 	"sort"
 	"strings"
 	"time"
@@ -528,6 +529,9 @@ func GetServiceLogs(ctx context.Context, serviceID string) (string, error) {
 
 	reader, err := client.ServiceLogs(ctx, serviceID, logOptions)
 	if err != nil {
+		if strings.Contains(strings.ToLower(err.Error()), "tty service logs only supported with --raw") {
+			return getServiceLogsRawCLI(ctx, serviceID)
+		}
 		return "", fmt.Errorf("failed to get service logs: %w", err)
 	}
 	defer func() {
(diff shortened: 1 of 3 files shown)


eldara-cruncher (commented):
## Review — Fix tty: true logs (#265)

**Approach is sound.** Falling back to `docker service logs --raw` via CLI when the SDK rejects TTY services is the right call — the SDK simply doesn't support this mode.

### Issues

**1. `io.ErrShortWrite` fallback is too broad** (`shouldFallbackToRawFromStdCopy`)

`io.ErrShortWrite` means "the writer accepted fewer bytes than requested" — it's a generic I/O error, not a TTY indicator. This could trigger the CLI fallback on unrelated write failures (e.g., pipe buffer pressure). I'd remove this case unless you've observed it specifically from TTY stdcopy failures.

**2. No tests**

`shouldFallbackToRawFromStdCopy` is a pure function — easy to table-test. Worth adding coverage given it controls a significant behavioral branch.

**3. Duplicated context + args construction**

Both `getServiceLogsRawCLI` (service.go) and `streamServiceLogsRawCLI` (stream.go) independently build docker CLI args with the same context-detection logic. Minor, but a shared helper (e.g. `buildDockerServiceLogsArgs(ctxName, serviceID string, extra ...string) []string`) would reduce the surface for divergence.

**4. Silent behavior change in stdcopy error path**

The original code logged `stdcopy finished with error` but did **not** send to the `errs` channel. This PR adds `errs <- scErr` in the non-fallback branch. That's arguably a bug fix, but it changes error propagation for all services — not just TTY ones. Worth calling out in the PR description.

**5. Error string matching is fragile (acknowledged)**

Matching on `"tty service logs only supported with --raw"` will break if Docker changes the message. No great alternative with the current SDK, but a comment noting this coupling would help future readers.

### Minor

- Default `bufio.Scanner` buffer is 64 KB. If a TTY service emits very long lines they'll be silently truncated/split. Probably fine in practice, but `sc.Buffer(make([]byte, 0, 256*1024), 256*1024)` is cheap insurance.

### Summary

The core fix is correct and well-structured. I'd want #1 (remove `ErrShortWrite`) and #2 (tests for `shouldFallbackToRawFromStdCopy`) addressed before merge. The rest are suggestions.

clangenb (approved):
Looks good tome!

PR #276, opened 2026-03-09
feat: migrate logs, inspect, help views to app-level search box

eldara-cruncher:
## Summary

Closes #275

- Implement `Filterable` interface on **logs**, **inspect**, and **help** views so `/` opens the unified app-level search box (filter-only — hides non-matching lines)
- Move the existing highlight+navigate search to **Ctrl+F** for logs and inspect views
- Add **n/N** match navigation to the inspect view (previously only logs had this)
- Fix pre-existing bug where `highlightContent()` in logs used raw-line indices instead of visible-line indices (caused wrong match positions when node filter was active)

## Test plan

- [x] `go build ./...` compiles
- [x] `go test ./views/logs/ ./views/inspect/ ./views/help/` — all pass (19 new tests added)
- [x] `golangci-lint run ./...` — 0 issues
- [x] Manual: **Logs** — `/` opens app search box → type "error" → lines filter → `Ctrl+F` → highlights within filtered lines → `n`/`N` navigate → Esc cascade works
- [x] Manual: **Inspect** — `/` filter → `Ctrl+F` search → match count in header → `n`/`N` cycle → works in YAML and raw mode
- [x] Manual: **Help** — `/` opens search → type "scale" → only matching keybindings/commands shown

### DIFF
--- a/views/help/help_test.go
+++ b/views/help/help_test.go
@@ -145,3 +145,71 @@ func TestView_EmptyCategories_CommandList(t *testing.T) {
 	// Should render command list path (even if empty)
 	require.Contains(t, out, "Help")
 }
+
+// --- Filterable (app-level "/" filter) tests ---
+
+func TestApplySearchQuery_CommandMode(t *testing.T) {
+	cmds := []CommandInfo{
+		{Name: "stacks", Description: "List stacks"},
+		{Name: "help", Description: "Show help"},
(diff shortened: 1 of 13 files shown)


clangenb (approved):
Looks good to me! Tested for logs and inspect view that first filtering and then search works, and that first search and then filtering works!

PR #278, opened 2026-03-10
fix: breadcrumb bar accumulates views unbounded

### LINKED ISSUE #255: [breadcrumbs] view stack accumulates views. It would be better to introduce top level views
If I am working in swarmcli and jumping from stacks to services logs etc the bottom line persist the history

Image

However I would like to see <stacks>  only on top level, services on the second etc.

eldara-cruncher:
## Summary

Fixes #255 — the bottom breadcrumb bar was showing the entire navigation history, growing without limit.

- **`IsTopLevel()` helper** in `views/view/constants.go` identifies root views (`stacks`, `nodes`, `configs`, `secrets`, `networks`, `contexts`, `loading`, `help`)
- **Stack trimming** in `viewstack.Push()`: trims to nearest top-level ancestor and caps at 10 entries
- **Breadcrumb rendering** extracted to pure `RenderBreadcrumbs()` in `app/breadcrumbs.go`: hides bar for top-level views, caps display at 3 items with `…` prefix when trimmed
- **Unit tests** for `IsTopLevel`, stack trim/cap, and all breadcrumb rendering edge cases

## Test plan

- [x] `go test ./views/view/... ./views/viewstack/... ./app/...` — all pass
- [x] `go build -v -o swarmcli .` — builds cleanly

### DIFF
--- a/app/breadcrumbs.go
+++ b/app/breadcrumbs.go
@@ -12,16 +12,17 @@ import (
 )
 
 // RenderBreadcrumbs produces the breadcrumb bar string from a list of view names.
-// Returns "" when the current (last) view is top-level.
-// Trims to the nearest top-level ancestor, then caps visible items at maxDisplay
-// (prepending "…" when items were trimmed).
+// Top-level views show only their own name. Nested views are trimmed to the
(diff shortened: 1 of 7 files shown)


clangenb (approved):
Tested, works!

PR #280, opened 2026-03-10
refactor: separate docker events from Bubble Tea TUI layer

### LINKED ISSUE #148: [docker] separate pure docker methods from TUI bubble tea functions
The docker domain becomes mixed purpose. Some functions are purely used for docker operations, where others return a tea.cmd function. The library is becoming complex to understand. We need a better separation of concerns for this.

eldara-cruncher:
## Summary
- Removed `bubbletea` dependency from the `docker/` package entirely
- Renamed `docker.EventMsg` → `docker.Event` and `docker.WatchEventsCmd() tea.Cmd` → `docker.WatchEvent() Event` (pure blocking function)
- Added `watchEventsCmd()` wrapper in `app/update.go` that wraps the pure call in a `tea.Cmd` closure — matching the pattern already used for every other docker operation

Closes #148

## Manual test: verify Docker events still refresh the UI

The event watcher drives live UI updates when Docker resources change. To verify:

1. **Start a Swarm** (`./test-setup/testenv.sh up && ./test-setup/testenv.sh deploy`)
2. **Run swarmcli** (`go run .`) — lands on the **Stacks** view
3. **In a separate terminal**, scale a service: `docker service scale <stack>_<service>=3`
4. **Observe the Stacks view** — the replica count should update automatically within a few seconds without pressing any key
5. **Navigate to the Nodes view**, then run `docker node update --label-add test=true <node-id>` — the nodes list should refresh automatically

If both views update reactively without manual refresh, the event watcher is working correctly.

### DIFF
--- a/app/model.go
+++ b/app/model.go
@@ -94,7 +94,7 @@ func (m *Model) Init() tea.Cmd {
 		loadSnapshotAsync(),
 		m.systemInfo.LoadStatus(),
 		m.systemInfo.Init(), // Initialize systeminfo's tick commands
-		docker.WatchEventsCmd(),
+		watchEventsCmd(),
 	)
 }
 
--- a/app/update.go
+++ b/app/update.go
@@ -28,7 +28,7 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	}
 
 	switch msg := msg.(type) {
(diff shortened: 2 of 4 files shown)


clangenb (approved):
Code looks good. Tested that services update!

PR #282, opened 2026-03-10
fix: app-level error dialog, remove error handling from loadingView

### LINKED ISSUE #52: Implement Error interface for views
The views cannot meaningfully handle errors yet. And error usually implies that the UI content fails to load.

### LINKED ISSUE #131: [loadingView] do not handle going to context view from there. it should be handled from the calling view.
This basically makes the loading view very nongeneric, as it assumes that we want to go to the context view for any error, which does not make sense to me. Also, it is bad practice that the loading view tries to infer implicitly based on the message if it supposed to show an error view instead as there is no separation of concerns.

The good pattern would be:
1. The calling view checks the message
2. The calling view then either invokes the loading view or the error dialog

eldara-cruncher:
## Summary

- **loadingView** stripped of all error handling — now a pure spinner. No more `strings.HasPrefix(message, "Error")` content sniffing or hardcoded "go to contexts view" navigation.
- **App-level error dialog** added to `app.Model`, reusing `confirmdialog` with `ErrorMode`. Any view can emit `view.AppErrorMsg` to trigger it; the app decides where to navigate on dismiss via `FallbackView`.
- **Tasks view** migrated from inline viewport error text to `AppErrorMsg` — dismiss goes back to the parent view.
- Snapshot load errors now show the app error dialog over the loading spinner, with contexts view as the fallback on dismiss (same UX, proper separation of concerns).

Closes #131, closes #52

## Test plan

- [x] `go build` / `go test ./...` / `golangci-lint` all pass
- [x] Start app with unreachable Docker host → error dialog appears over loading spinner → Enter navigates to contexts view
- [x] Switch to invalid context → loading spinner → error dialog → Enter → contexts view
- [x] Normal startup → loading spinner → stacks view (no regression)
- [ ] Navigate to tasks of a service that fails to load → error dialog → dismiss returns to previous view

### DIFF
--- a/app/model.go
+++ b/app/model.go
@@ -6,6 +6,7 @@ package app
 import (
 	"swarmcli/docker"
 	"swarmcli/views/commandinput"
+	"swarmcli/views/confirmdialog"
 	loadingview "swarmcli/views/loading"
 	"swarmcli/views/searchinput"
 	systeminfoview "swarmcli/views/systeminfo"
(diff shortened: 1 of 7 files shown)


clangenb (commented):
Works!

clangenb (approved):

PR #291, opened 2026-03-14
Add versioncheck

### LINKED ISSUE #198: Check if there's a newer version
Check automatically if we released a newer version and show it in the header.

mosonyi:
Fixes #198


eldara-cruncher (commented):
Here's my review of this PR.

**What it does:** On startup, sends a POST to `https://swarmcli.io/api/v1/version` with the current version and edition, then displays a hint (e.g. `⚡v1.3.3`) in the system info panel if a newer release exists. Also threads `edition` (ce/be) through main → app → views.

**Overall:** Solid implementation — clean separation, good test coverage with httptest, version comparison logic is correct. A few issues worth addressing:

### Bugs / correctness

1. **`LatestVersionMsg` returned as `nil` is swallowed silently by Bubble Tea, but `shouldShowLatestVersion` returning false causes `return nil` inside the `tea.Cmd` func.** Returning `nil` from a `tea.Cmd` is valid but it's a `tea.Msg(nil)` — if any other `Update` handler has a `case nil:` or a default branch that processes unexpected messages, this could trigger unintended behavior. Consider returning a dedicated "no update needed" message type, or simply not returning a command at all. At minimum, document this is intentional.

2. **`versionCheckURL` is a mutable package-level var** (`views/systeminfo/model.go`). Tests mutate it with `t.Cleanup` to restore, but if tests run in parallel (`t.Parallel()`), this is a data race. Currently tests don't use `t.Parallel()` so it works, but it's fragile. Consider accepting the URL via the `Model` constructor or using a `checkURL` field on the struct.

### Security / privacy

3. **The POST to `swarmcli.io/api/v1/version` sends the current version and edition on every startup with no opt-out.** This is essentially a phone-home. Users should have a way to disable it (env var, config flag). At minimum, document it. Also consider: is the server logging IPs? This could be a concern for enterprise users of the "be" edition.

4. **No TLS certificate pinning or verification override** — this is fine (Go's default `http.Client` verifies TLS), just noting it's correctly using defaults.

### Design

5. **4-second timeout (`versionCheckTimeout`) on the critical startup path.** If the server is slow or unreachable, the UI still renders (it's async via `tea.Cmd`), but the goroutine hangs for up to 4s. This is acceptable since it doesn't block the UI, but worth confirming that's the intended behavior. A context with cancellation on app shutdown would be cleaner.

6. **`editionLabel` in `app.go` handles `""` both in the `switch` default and explicitly in `SetEdition`.** The empty-string case in `editionLabel` is dead code since `SetEdition` already normalizes to `"ce"` before calling it.

7. **`helpbar.EditionLabel` is set as a side effect of `SetEdition`** — this creates a hidden coupling between `app` and `views/helpbar`. The diff doesn't show the `helpbar` package, so I can't verify thread safety of that assignment.

### Nits

8. **`message` field on `LatestVersionMsg` is stored but never displayed.** The server returns a message string, the model stores it, the test asserts it, but `versionDisplay()` never renders it. Either display it or drop it (YAGNI). If it's reserved for future use, that's speculative code.

9. **Pre-release versions (e.g. `1.3.0-rc.1`) are stripped to `1.3.0` by `parseVersion`.** This means a user on `1.3.0-rc.1` would not be prompted to upgrade to `1.3.0` stable (they'd compare equal). This may or may not be desired — worth a conscious decision.

### Tests

Tests are well-structured — httptest usage, cleanup of global state, edge cases (dev build, same version, server error, edition override). Good coverage. The `TestIsNewerVersion` cases are clear and correct.

---

**Verdict:** Approve with minor changes. Items 1, 6, and 8 are easy fixes. Item 3 (opt-out for phone-home) is the most important to address before merge, especially for enterprise users. Item 2 is worth fixing to avoid future test flakiness.

PR #295, opened 2026-03-16
Deduplicate task error helpers, add cutoff and tests

eldara-cruncher:
## Summary
- Extracts `taskKeyForService`, `latestTasksByServiceKey`, and `activeDeploymentErrorsByService` into a shared `views/taskutil` package, eliminating identical copies across `views/services` and `views/stacks`
- Adds 5-minute time cutoff to `activeDeploymentErrorsByService` (matching existing error detection logic) to avoid surfacing stale historical failures
- Handles empty `errMsg` on failed/rejected tasks by falling back to `"task <state>"` instead of leaving error text blank
- When multiple slots error for the same service, picks the most recent error deterministically
- Adds unit tests covering: slot key generation, timestamp fallback, running-task preference, failed rolling update detection, recovered service clearing, stale error filtering, empty error message fallback, and rejected task handling

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./views/taskutil/` — 8 tests pass
- [x] Manual verification on a swarm cluster with a failing rolling update

### DIFF
--- a/views/services/update.go
+++ b/views/services/update.go
@@ -16,6 +16,7 @@ import (
 	inspectview "swarmcli/views/inspect"
 	logsview "swarmcli/views/logs"
 	"swarmcli/views/scaledialog"
+	"swarmcli/views/taskutil"
 	"swarmcli/views/view"
 	"time"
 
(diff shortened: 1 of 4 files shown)


mosonyi (approved):
I checked manually. All good!

PR #298, opened 2026-03-17
fix: refresh expanded tasks from fresh snapshot

eldara-cruncher:
## Summary

Fixes #284 — tasks in the services view now refresh on the fly while expanded.

### Root cause

`refreshExpandedTasksCmd` and `checkServicesCmd` are dispatched concurrently from the `TickMsg` handler (every 2s). `checkServicesCmd` calls `TriggerRefreshIfNeeded()` which starts an **async** (non-blocking) background goroutine to refresh the snapshot. Meanwhile, `refreshExpandedTasksCmd` calls `GetTasksForService()` → `GetSnapshot()` which returns the **stale** cached snapshot — the background refresh hasn't completed yet. On the next tick, the snapshot TTL (3s) often hasn't expired, so no new refresh is triggered. The result: tasks are perpetually read from stale data.

Collapsing and re-expanding works because the manual toggle happens at a different point in the refresh cycle — typically after the background refresh has completed — so `GetSnapshot()` returns fresh data.

### Fix

Replace the `tea.Batch` of per-service closures with a single `tea.Cmd` that calls `GetOrRefreshSnapshot()` (blocking, synchronous) **before** fetching tasks. This guarantees the global snapshot cache is fresh when `GetTasksForService()` reads from it. The blocking call runs in a goroutine (`tea.Cmd`), so it does not freeze the UI. When the snapshot is already fresh (within 3s TTL), `GetOrRefreshSnapshot()` returns instantly from cache.

### Flow after fix

```
TickMsg (every 2s)
  → checkServicesCmd goroutine (unchanged, async)
  → refreshExpandedTasksCmd goroutine:
      1. GetOrRefreshSnapshot()   ← blocks ~0.5s if stale, instant if fresh
      2. GetTasksForService(sid)  ← reads FRESH snapshot
      3. returns AllTasksLoadedMsg
  → Update() stores fresh tasks in m.serviceTasks
  → View() re-renders with updated task rows
```

## Test plan

- [x] Unit tests pass (`go test ./views/services/...`)
- [x] Lint passes (`golangci-lint run ./...`)
- [x] Build passes
- [ ] Manual: expand a service with `p`, scale/restart it, verify tasks update without collapsing

### DIFF
--- a/views/services/messages.go
+++ b/views/services/messages.go
@@ -54,3 +54,8 @@ type TasksLoadedMsg struct {
 	ServiceID string
 	Tasks     []docker.TaskEntry
 }
+
+// AllTasksLoadedMsg carries refreshed tasks for all expanded services at once.
+type AllTasksLoadedMsg struct {
+	Tasks map[string][]docker.TaskEntry
+}
--- a/views/services/model.go
+++ b/views/services/model.go
@@ -173,7 +173,7 @@ func (m *Model) ShortHelpItems() []helpbar.HelpEntry {
 }
 
 func (m *Model) OnEnter() tea.Cmd {
-	return nil
+	return tickCmd()
 }
 
 func (m *Model) OnExit() tea.Cmd {
--- a/views/services/model_test.go
+++ b/views/services/model_test.go
@@ -302,6 +302,21 @@ func TestTasksLoadedMsg_StoresTasks(t *testing.T) {
 	require.Len(t, m.serviceTasks["id-web"], 1)
 }
 
+func TestAllTasksLoadedMsg_StoresAllTasks(t *testing.T) {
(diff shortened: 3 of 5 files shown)


mosonyi (changes requested):
I did a manual test. 
What I have changed: 
1. I wanted to do a rolling update to a version that does not exist.
2. I pressed 'p' to see if it tried to coninuously download -> OK
3. Pressed 'l' to see the logs -> OK
4. Pressed ESC to go back
5. The processes are not updating anymore automatically -> NOT OK!
6. When I press 'p' 2 times, it updates for that particular moment, but no continuous updates anymore.

mosonyi (approved):
The manual test worked!

PR #302, opened 2026-03-19
fix: forward `/` and `:` to active inputs

eldara-cruncher:
## Summary
- Forward `/` and `:` as text to commandInput, searchInput, and view-internal search (ctrl+f) instead of swallowing them or triggering actions
- Fixes 5 related bugs in `app/update.go` where the `:` and `/` handlers ran before the input-forwarding logic
- Adds tests for all 5 cases

Fixes #300

## Test plan
- [ ] `go test ./app/... -run "TestSlash|TestColon"` — 5 new tests pass
- [ ] Manual: open search with `/`, type `/` and `:` — characters appear
- [ ] Manual: open command with `:`, type `:` and `/` — characters appear
- [ ] Manual: in logsview ctrl+f search, type `:` — character appears

### DIFF
--- a/app/update.go
+++ b/app/update.go
@@ -126,6 +126,24 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				}
 			}
 
+			// If command input is already visible, forward ":" as text
+			if m.commandInput.Visible() {
+				cmd := m.commandInput.Update(msg)
+				return m, cmd
+			}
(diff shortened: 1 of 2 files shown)


mosonyi (changes requested):
In log view, I can search now for '/'.
I don't like the idea of searching for something, then all the other logs are hidden.
Better to show everything but highlight the search text.

mosonyi (approved):
Manual test: OK!

PR #306, opened 2026-03-24
fix: populate Name in snapshot hash to prevent cursor jumping

eldara-cruncher:
## Summary
- The `stableSecret`/`stableConfig` structs in `secretsLoadedMsg`/`configsLoadedMsg` handlers declared a `Name` field but never populated it, while `checkSecretsCmd`/`checkConfigsCmd` did populate it
- This hash mismatch caused every poll cycle to detect a false "change" and trigger a full reload, making the cursor jump back and forth after sorting
- Added regression tests that verify the hash from the load handler matches the hash from the poll checker for identical data

## Test plan
- [x] `go test ./views/secrets/... ./views/configs/...` passes
- [ ] Manual: open secrets view, sort by any column, confirm cursor stays stable (no jumping every few seconds)
- [ ] Check debug log for `checkSecretsCmd: No changes detected` instead of repeated `Change detected! Refreshing secret list`

### DIFF
--- a/views/configs/actions_test.go
+++ b/views/configs/actions_test.go
@@ -199,6 +199,31 @@ func TestCheckConfigsCmd_NoChange_ReturnsPollRetry(t *testing.T) {
 	require.True(t, isPollRetry, "should return PollRetryMsg when no change")
 }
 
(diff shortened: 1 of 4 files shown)


mosonyi (approved):
Manual test ok!

PR #314, opened 2026-03-26
fix: skip version check for dev builds

### LINKED ISSUE #198: Check if there's a newer version

eldara-cruncher:
## Summary
- Skip the remote version check entirely when the current version is `"dev"` (unbuilt/local binary), avoiding unnecessary API calls during development.

Closes #198

## Test plan
- [x] `TestCheckLatestVersion_DevBuildSkipsCheck` asserts `nil` return (no command issued)
- [x] All existing version-check tests pass

### DIFF
--- a/views/systeminfo/model.go
+++ b/views/systeminfo/model.go
@@ -112,6 +112,11 @@ func (m *Model) CheckLatestVersion() tea.Cmd {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


clangenb (approved):
Manual test was ok.

PR #318, opened 2026-03-27
fix: skip stack redeploy when editor content is unchanged

eldara-cruncher:
## Summary
- Skip redeployment when the user exits the stack editor without making changes
- Carries original YAML through `editorContentMsg` and compares before deploying
- Follows the same pattern already used in the configs edit flow

Closes #316

## Test plan
- [x] New unit test `TestUpdate_EditorContentMsg_EditMode_NoChange` asserts no deploy when content is unchanged
- [x] Existing `TestUpdate_EditorContentMsg_EditMode` updated to verify deploy still happens on actual changes
- [ ] Manual: press `e` on a stack, exit editor without changes — stack should not redeploy


mosonyi (approved):
LGTM

PR #319, opened 2026-03-27
fix: show per-service secrets and configs in stack inspect

eldara-cruncher:
## Summary
- Add `Secrets` and `Configs` fields to `ServiceSummary` so the stack inspect view shows which secrets/configs each service references
- Add a test secret to the demo stack and extend integration test assertions for secrets and configs

Closes #317

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] `golangci-lint run ./... --build-tags=integration` — 0 issues
- [ ] `./test-setup/testenv.sh integration` — verify secrets/configs appear in inspect output


mosonyi (approved):
LGTM

PR #321, opened 2026-03-30
fix: add context timeouts to all Docker API calls and prevent polling goroutine pile-up

eldara-cruncher:
## Summary

- Add `ctx context.Context` parameter to all Docker-layer functions that
  previously created their own `context.Background()` (ServiceOps, StackOps,
  NodeOps, ContextOps interfaces)
- Add context timeouts to all view-layer Docker API calls: 4s for background
  polling, 15s for user-initiated operations
- Add `atomic.Bool` in-flight guard to prevent overlapping `checkConfigsCmd` /
  `checkSecretsCmd` goroutines
- Add unit tests for timeout/unreachability handling and AST-based guardrail
  test to prevent regression

## Problem

When Docker becomes unreachable, the configs/secrets views freeze and become
unresponsive. All Docker API calls used `context.Background()` with no timeout,
so each call hung for the OS TCP timeout (30-60s). Meanwhile, the 5-second
polling tick kept spawning new goroutines — after 30 seconds, 6+ goroutines
were simultaneously blocked, each making N+1 API calls (1 list + N inspects).

**Why this was only observed on BE, not CE:** the bug exists in both editions
(same polling code), but the failure mode differs by connection type. CE users
typically connect via unix socket — when the daemon is down, the kernel returns
ECONNREFUSED instantly, so polls fail fast with no pile-up. BE requires TCP
connections (the shell feature derives a WebSocket URL from the Docker host and
rejects unix sockets), so all BE deployments use TCP contexts. When a TCP host
becomes unreachable there is no immediate rejection — the kernel enters TCP
retransmit and hangs for 30-60 seconds, which is what causes the goroutine
pile-up.

This is a codebase-wide fix — the networks view already followed the timeout
pattern correctly, but configs, secrets, nodes, services, stacks, and contexts
did not.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (includes new timeout/unreachability unit tests)
- [x] `golangci-lint run ./...`
- [ ] `./test-setup/testenv.sh integration`
- [ ] Manual: connect to Swarm via TCP context, navigate to configs, kill
      Docker daemon — app should remain responsive (polls fail within 4s)

### DIFF
--- a/docker/context.go
+++ b/docker/context.go
@@ -150,7 +150,7 @@ func UseContext(contextName string) error {
 }
 
 // ValidateContext checks if a context switch would succeed by attempting to connect
-func ValidateContext(contextName string) error {
+func ValidateContext(ctx context.Context, contextName string) error {
 	// Save current context
 	currentCtx, err := GetCurrentContext()
 	if err != nil {
@@ -170,7 +170,6 @@ func ValidateContext(contextName string) error {
 		return fmt.Errorf("failed to connect to context %s: %w", contextName, err)
 	}
 	// Verify connection with ping
-	ctx := context.Background()
 	if _, err := cli.Ping(ctx); err != nil {
 		// Switch back to original context
 		_ = UseContext(currentCtx)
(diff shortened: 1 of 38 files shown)


clangenb (approved):
Refactoring changes look solid to me. I will merge it. It is definitely an improvement if the responsiveness issue is still not completely fixed, it is probably due to another reason.

PR #331, opened 2026-04-15
fix: handle pgup/pgdown in contexts view

### LINKED ISSUE #329: pg+up/down at context
At context view the pg up and down doesn't work. any other views are ok.

eldara-cruncher:
## Summary

- `pgup`/`pgdown` were missing from the keyboard `switch` in `views/contexts/update.go`
- All other list views (`stacks`, `services`, `nodes`) handle these keys; contexts was the odd one out
- Added two cases that call `m.MoveCursor(±m.List.Viewport.Height)`, identical in shape to the existing `up`/`down` cases
- Added `TestKey_PgUpPgDown` unit test

Closes #329

## Test plan
- [ ] `go test ./views/contexts/...` passes (including `TestKey_PgUpPgDown`)
- [ ] Manual: open contexts view with >1 page of contexts, verify PgUp/PgDown scrolls

### DIFF
--- a/views/contexts/update.go
+++ b/views/contexts/update.go
@@ -692,6 +692,14 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 			m.MoveCursor(1)
 			return nil
 
+		case "pgup":
+			m.MoveCursor(-m.List.Viewport.Height)
+			return nil
+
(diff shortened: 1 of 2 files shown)


mosonyi (approved):
LGTM

PR #382, opened 2026-06-03
feat(helpbar): add 'ctrl+q' shortcut for quitting in short help items

Arka-cell:
Still getting used to the codebase and the CLI. Add missing short key to quit.

Using `ctrl+q` for `swarmcli` with VSCode Terminals results in conflict between a VSCode shortcut key. I think using `ctrl+c` or `q` could be better to aim for simplicity.

### DIFF
--- a/app/view.go
+++ b/app/view.go
@@ -28,6 +28,7 @@ func (m *Model) View() string {
 	globalHelp := []helpbar.HelpEntry{
 		{Key: "f", Desc: "Fullscreen"},
 		{Key: "?", Desc: "Help"},
+		{Key: "ctrl+q", Desc: "Quit"},
 	}
 	if m.currentView.Name() == view.NameHelp {
 		globalHelp = []helpbar.HelpEntry{}
--- a/views/stacks/model.go
+++ b/views/stacks/model.go
@@ -209,7 +209,6 @@ func (m *Model) ShortHelpItems() []helpbar.HelpEntry {
 		{Key: "↑/↓", Desc: "Navigate"},
 		{Key: "/", Desc: "Filter"},
 		{Key: "?", Desc: "Help"},
-		{Key: "ctrl+q", Desc: "Quit"},
 	}
 }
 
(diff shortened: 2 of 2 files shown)


clangenb (changes requested):
Thanks for the PR. The binding you picked is correct — no need to second-guess it. ctrl+q is the canonical quit key (helpbar.KeyQuit), handled in app/update.go:

```go
  // Global quit handler: Ctrl+Q (primary) and Ctrl+C (standard terminal fallback)
  if msg.String() == "ctrl+q" || msg.Type == tea.KeyCtrlC {
      exitCmd := m.currentView.OnExit()
      return m, tea.Batch(exitCmd, tea.Quit)
  }
```
  ctrl+c routes through the same handler, so both keys quit identically (run the current view's OnExit, then tea.Quit). We treat ctrl+q as the primary and intentionally don't advertise ctrl+c — it's the standard terminal-interrupt fallback, and surfacing both in the help bar would just add noise. So ctrl+q is the right thing to document.

  The change I'd ask for is the placement. Quit works on every screen, so it belongs in the global help bar rather than a single view's ShortHelpItems(). As written, the hint only shows on the stacks view; the other ~11 views (services, nodes, secrets, logs, …) still wouldn't show it, which is inconsistent.

  The global help is assembled in app/view.go:
```go
  globalHelp := []helpbar.HelpEntry{
      {Key: "f", Desc: "Fullscreen"},
      {Key: helpbar.KeyQuit, Desc: "Quit"},  // add here
      {Key: "?", Desc: "Help"},
  }
```
  Adding it there surfaces Quit on every view in one line, and using helpbar.KeyQuit keeps it tied to the source of truth. The addition in views/stacks/model.go can then be dropped.

  This also closes a real gap: quit currently isn't shown anywhere in the live help bar, so the improvement is worth having once it's in the global spot.

clangenb (approved):
Thanks for the fix. Looks good to me!

PR #384, opened 2026-06-03
Feat  editor-fallback

Arka-cell:
Compose to Swarm Stack conversion results in the following error:

```bash
⚠ editor failed: exec: "nano": executable file not found in %PATH%  
```

To mitigate this error, I updated the default editor fallback logic.

In the codebase, when the EDITOR environment variable is not defined, the application was hardcoded to fall back to "nano". Since Windows does not have nano installed by default, running operations that spawn an editor (such as configuring services, stacks, or secrets) fails with the error you saw: `exec: "nano": executable file not found in %PATH%.`

### What I did to fix it:
I modified the editor selection logic in the following files:

```
views/stacks/editcmd.go
views/configs/editcmd.go
views/secrets/actions.go
```

For each file, the logic has been updated to check the OS environment using Go's standard runtime package:

It first looks up the EDITOR environment variable.
If empty, it checks the standard fallback VISUAL.
If both are empty, it checks if the current OS is Windows (`runtime.GOOS == "windows"`). If so, it falls back to the built-in GUI editor notepad. Otherwise, it defaults to nano.

Here is an example of the updated logic now in place:

```go
	editor := os.Getenv("EDITOR")
	if editor == "" {
		if runtime.GOOS == "windows" {
			editor = "notepad"
		} else {
			editor = "nano"
		}
	}
```
Since notepad is standard on every Windows machine, SwarmCLI will now launch Notepad when you trigger stack edit actions, allowing you to edit the compose templates smoothly without any extra configuration.

### DIFF
--- a/core/primitives/editor/editor.go
+++ b/core/primitives/editor/editor.go
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package editor
+
+import (
+	"os"
+	"runtime"
+)
+
+// Sense returns the preferred text editor for the current system.
+// It checks the EDITOR env var, falls back to VISUAL, and finally
+// defaults to notepad on Windows or nano on other platforms.
+func Sense() string {
+	editor := os.Getenv("EDITOR")
+	if editor == "" {
+		editor = os.Getenv("VISUAL")
+	}
+	if editor == "" {
+		if runtime.GOOS == "windows" {
+			editor = "notepad"
+		} else {
+			editor = "nano"
+		}
+	}
+	return editor
+}
--- a/core/primitives/editor/editor_test.go
+++ b/core/primitives/editor/editor_test.go
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package editor
+
+import (
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSense_EditorSet(t *testing.T) {
+	t.Setenv("EDITOR", "my-custom-editor")
+	t.Setenv("VISUAL", "my-visual-editor")
+
+	require.Equal(t, "my-custom-editor", Sense())
+}
+
+func TestSense_VisualSet(t *testing.T) {
+	t.Setenv("EDITOR", "")
+	t.Setenv("VISUAL", "my-visual-editor")
+
+	require.Equal(t, "my-visual-editor", Sense())
+}
+
+func TestSense_Default(t *testing.T) {
+	t.Setenv("EDITOR", "")
+	t.Setenv("VISUAL", "")
+
+	expected := "nano"
+	if runtime.GOOS == "windows" {
+		expected = "notepad"
+	}
+
+	require.Equal(t, expected, Sense())
+}
--- a/views/configs/editcmd.go
+++ b/views/configs/editcmd.go
@@ -44,9 +44,6 @@ func editWithTempFileCmd(baseName string, initialData []byte, onDone func([]byte
 
 	editor := os.Getenv("EDITOR")
 	if editor == "" {
-		editor = os.Getenv("VISUAL")
-	}
-	if editor == "" {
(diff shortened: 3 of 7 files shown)


clangenb:
Thanks for tackling the Windows editor issue! A few things to address before merge:

  1. views/configs/editcmd.go is missing the VISUAL fallback. This file only got the runtime.GOOS == "windows" branch — it doesn't add the os.Getenv("VISUAL") step that stacks/editcmd.go and secrets/actions.go received. The PR description says VISUAL is checked everywhere, so this looks like a copy-paste miss. It now behaves differently from its siblings:

```go
  editor := os.Getenv("EDITOR")
  if editor == "" {
      editor = os.Getenv("VISUAL") // <-- missing here
  }
  if editor == "" {
      if runtime.GOOS == "windows" {
          editor = "notepad"
      } else {
          editor = "nano"
      }
  }
```

  2. The resolution logic is now duplicated across three files (and already drifting). With EDITOR → VISUAL → OS-default, this is non-trivial enough to warrant a single helper rather than three copies. Suggest extracting it to a shared package (e.g. core/primitives/editor, next to the existing hash package) and calling it from all three sites. This PR already touches all
  three, so it's a good moment to unify.

  3. No tests. Once extracted, the helper is straightforward to cover with a table-driven test: EDITOR set / VISUAL set / both empty, crossed with runtime.GOOS being windows vs. otherwise.

  4. views/secrets/actions.go doesn't split editor args. It calls exec.Command(editor, tmp.Name()) directly, whereas stacks and configs do strings.Fields(editor) first. So EDITOR="code --wait" works in stacks/configs but breaks in secrets. Pre-existing, but a shared helper (point 2) would fix it for free.

Nitpick: I think the title is wrong, the change is editor-fallback, not compose-conversion-specific.

Arka-cell:
@clangenb followed with your proposition (Sorry about the branch title I couldn't rename it maybe due to the fork still investigating) - help appreciated if you have an idea on how to do it

clangenb:
Thanks a lot, this looks almost good.

  Thanks for this — the Windows/VISUAL fallback and the Sense() extraction are clean, and pulling the three editor call sites onto one helper is a nice consolidation. Two things before merge:

  1. CI lint will fail (must fix). The new secureWipeAndRemove and its tests drop the _ = on a few cleanup calls, which trips errcheck — and golangci-lint run ./... is a hard gate in CI (ci.yml). Four spots:
  - views/secrets/actions.go:412 → _ = os.Remove(name)
  - views/secrets/actions.go:418 → _ = file.Close()
  - views/secrets/actions_test.go:229 and :248 → _ = os.Remove(tmp.Name())

  The code you replaced already used _ = os.Remove(...), so this just restores that convention.

  2. Let's drop the zero-overwrite entirely. I'd remove secureWipeAndRemove and go back to a plain os.Remove. The intent is good, but the overwrite only meaningfully defeats undelete on a traditional non-COW filesystem on real media. In the setups we actually run on it does little to nothing:
  - os.CreateTemp("") lands in /tmp, which is tmpfs (RAM) on most modern Linux — nothing persistent to wipe.
  - On SSDs, wear-leveling writes the zeroes to a different physical cell and leaves the original holding plaintext until GC.
  - On COW filesystems (btrfs/ZFS/APFS, and the overlayfs Docker Desktop uses), the overwrite allocates new extents; the old plaintext extent survives (and snapshots pin it).

  It's also only one of several plaintext copies — the editor itself can leave .swp/backup/undo files plus process memory. So this is a guarantee we can't actually keep, and a "secure wipe" in the code reads as stronger than it is. No defense-in-depth is better than a false sense of security here. If we do want to harden this later, the honest fix is to keep the secret off
  persistent disk in the first place (e.g. an explicitly tmpfs-backed temp dir), not to overwrite after the fact.

  That also removes the only piece of this PR that needs the extra test, keeping the change nicely surgical.
  
  Regarding the branch name: You cannot change a remote branch name, only create a different one. However, I do not care about your branch's name, only the PR title/description is relevant information to me. :)

clangenb:
Hey there, are you going to continue working on that? I think it is a useful feature. :)

clangenb:
Hey there, I will close this PR for now, as it is stale. Please just re-open it, when you feel like working on it again.

PR #386, opened 2026-06-04
feat(docker): add GetLocalNodeID accessor for the connected swarm node

eldara-cruncher:
## What

Adds `docker.GetLocalNodeID(ctx)` — returns the swarm node ID of the daemon the active Docker client is connected to (`cli.Info(ctx).Swarm.NodeID`), or `""` if it is not an active swarm node.

## Why

A small, generic extension point next to the existing node helpers in `docker/node.go`. It lets a caller learn which node it is talking to — e.g. to pin a short-lived workload to the same daemon it will read results from, so it doesn't depend on cross-node behaviour.

## Notes

No behaviour change. Additive accessor only; mirrors the `cli.Info` usage already in `docker/context.go`.

### DIFF
--- a/docker/node.go
+++ b/docker/node.go
@@ -10,6 +10,21 @@ import (
 	"github.com/docker/docker/api/types/swarm"
 )
 
+// GetLocalNodeID returns the swarm node ID of the daemon the active Docker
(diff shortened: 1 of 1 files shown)


mosonyi:
LGTM

PR #390, opened 2026-06-05
feat(logs): hide stopped-container logs by default (#388)

### LINKED ISSUE #388: [LOG] Don't show stopped container logs
With the tail log view, I can see the already stopped containers, aka old logs as well.
My suggestion is that by default, do not show them.

eldara-cruncher:
## Summary

Closes #388. The log view streams `docker service logs`, which the Docker API aggregates across **all** of a service's tasks — including the stopped containers of previous deployments. After a redeploy this mixes old/dead logs into the tail view. This hides stopped-task logs **by default**, with a `t` toggle to bring them back.

Each log line already carries `com.docker.swarm.task.id` (the stream sets `Details: true`), so we plumb the full task ID per line and hide lines whose task is no longer `DesiredState == Running`, derived from the snapshot **at render time** (so it tracks current cluster state). This matches the existing `extractUniqueNodes` convention.

## Behaviour

- Default: only currently-running tasks' lines are shown. FrameTitle reads `Stopped: hidden`.
- Press `t` to toggle → old/stopped lines appear, title reads `Stopped: shown`.
- Lines with **no parseable task ID** (tty/raw service logs, `Error:` lines, `--- stream closed ---`) always stay visible.
- **Fail-open:** when no snapshot is available yet, all lines are shown (no crash on startup).

## Implementation notes

- `lineTasks []string` parallel slice tracks the full task ID per line, kept in lock-step with `lines`/`lineNodes` on append, MaxLines trim, stream err/done, and `SetContent` (which also resets the previously-stale `lineNodes`).
- A single `lineVisible()` predicate is shared by `buildContent()` and `highlightContent()` so the node filter, `/` text filter, and hide-stopped filter compose, and `n`/`N` search-match indices stay aligned with the rendered output.

## Testing

`go build ./...`, `go vet`, `gofmt -l`, `golangci-lint run` (0 issues), full `go test ./...`, and `go test -race ./views/logs/...` all pass. Added 9 new unit tests (default state, toggle key, hide/show, fail-open, empty-task-ID always visible, search-index sync, SetContent reset) and updated existing tests for the 3-field marker / title / help changes.

### DIFF
--- a/views/logs/handlekey.go
+++ b/views/logs/handlekey.go
@@ -236,6 +236,14 @@ func HandleKey(m *Model, k tea.KeyMsg) tea.Cmd {
 			l().Infof("[logsview] 'o' key pressed: showing node selection dialog with %d nodes", len(nodes))
 		}
 		return nil
+	case "t":
+		// Toggle showing logs from stopped (non-running) tasks
+		old := m.getHideStopped()
+		m.setHideStopped(!old)
+		l().Infof("[logsview] 't' key pressed: hideStopped %v -> %v", old, !old)
+		return func() tea.Msg {
+			return HideStoppedToggledMsg{}
+		}
 	}
 
 	return nil
--- a/views/logs/logs_test.go
+++ b/views/logs/logs_test.go
@@ -11,6 +11,7 @@ import (
 	"swarmcli/docker"
 
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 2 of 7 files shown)


mosonyi:
Looking good. Now try to show older logs. For instance if I press p (to show even the shutdown state) and if I choose it, try to get the latest log from there.

image

PR #393, opened 2026-06-06
fix(services): content-aware column widths, drop redundant STACK (#392)

### LINKED ISSUE #392: Better alignment in services
As you can see, the service name is not visible fully; we have enough space when there's no error.

My suggestions are:
* Remove stack (we know already that where we are)
* Try to remove the empty space in favour of longer columns
* If we can't do it because the screen is small and we should use the ... (as the current state), leave at least one space between the 2 columns.

Image

eldara-cruncher:
Fixes #392.

## Problem
The services view split the viewport into 10 **equal** column slices regardless of content, so the SERVICE name was truncated with `…` even on a wide terminal while STATUS/MODE/REPLICAS and an empty ERROR column sat half-empty. Columns also abutted with no separating space, and STACK was shown even when every row had the same stack.

## Change
- **Single source of truth for columns.** Replaced the four parallel hardcoded lists (header labels, the sort-index map, the width array, the row format specifiers) with one ordered `[]colSpec` (`views/services/columns.go`) carrying each column's label, sort field, width policy, and cell extractor. Header labels, sort indices, widths, and row cells all derive from it, so they can't drift out of sync.
- **Content-aware widths.** Each column sizes to the widest of its header label and any visible cell; a 1-space inter-column gap is baked into every width; leftover slack goes to the flex columns (SERVICE/IMAGE/PORTS/ERROR), and on overflow flex columns shrink first with floors that never drop below the (untruncated) header label — so the header stays aligned with rows and the gap always survives. The sorted column reserves room for its ` ▲`/` ▼` indicator.
- **STACK dropped when redundant.** Removed only outside the *All Services* scope (single stack/node), where every row carries the same value. Kept in *All Services* where it varies — see note below.
- **Rune-aware truncation.** `truncateWithEllipsis` now counts/slices by runes so width measurement and output agree.
- Selected-row horizontal scroll, full-width highlight, red error rows, and the expanded task sub-rows are unchanged.

## Note on the issue
The issue asked to "remove the stack column". I scoped that to *redundant* cases only: in the *All Services* view STACK is the only per-row stack indicator and genuinely varies, so removing it there would regress that view. It's dropped in single-stack/node scopes (the reporter's context) and kept in All-Services.

## Tests
New `views/services/columns_test.go`: STACK presence per scope, header/width length sync, full SERVICE name on a wide terminal (#392 regression), label-width floor preserved when narrow, empty/zero-width safety, sort-index resolution across the STACK shift, **header↔row width alignment across scopes/sort-fields/widths**, and rune-aware truncation. `go test ./...`, `go vet`, and `golangci-lint run ./views/services/...` all pass.

### DIFF
--- a/views/services/columns.go
+++ b/views/services/columns.go
@@ -1,0 +1,222 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package servicesview
+
+import (
+	"fmt"
+	"unicode/utf8"
+
+	"swarmcli/docker"
+	filterlist "swarmcli/ui/components/filterable/list"
+)
+
+// colID identifies a logical services column independent of its position in the
+// active column set (the STACK column is dropped outside the All-Services scope).
+type colID int
+
+const (
+	colService colID = iota
+	colStack
+	colReplicas
+	colStatus
+	colMode
+	colImage
+	colPorts
+	colCreated
+	colUpdated
+	colError
+)
+
+// colSpec is the single source of truth for a services column: its header label,
+// width policy, sort mapping, and how to extract its cell text from an entry.
+// Header labels, sort indices, column widths, and row cells all derive from the
+// active []colSpec so they can never drift out of sync.
+type colSpec struct {
+	id       colID
+	label    string
+	minWidth int  // content floor (excludes the inter-column gap)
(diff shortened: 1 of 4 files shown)


mosonyi (approved):
Perfect!

PR #395, opened 2026-06-06
feat(ui): shared content-aware column layout for list views (#394)

### LINKED ISSUE #394: Better alignment in secret and config view
Based on the fix of the service view, here would be nice to have the same logic.

1. If the first column is too long, add 1 more space between the 2 columns
2. With the right and left arrow, I should be able to move the unseen parts (similar to services)
3. Better alignment when there are free spaces in-between

Image

eldara-cruncher:
Fixes #394.

## Problem
Issue #394 asks for the services column fix (#392) on the secrets and config views: (1) keep ≥1 space after a long first column, (2) left/right arrows scroll the unseen parts of truncated cells, (3) content-aware widths that reclaim free space. The secrets, configs **and** nodes views all shared the same defects — generic equal-division widths, byte-based truncation, and an ad-hoc scroll that only moved the LABELS column.

## Change
Rather than copy the services logic into each view, the algorithm is extracted into a **reusable generic layer** in the filterable-list component and adopted by all four views (you chose this scope: secrets + configs + nodes, plus migrating services so there's a single implementation).

**New `ui/components/filterable/list/layout.go`:**
- generic `Column[T]{Label, MinWidth, Flex, Cell}`
- `LayoutWidths` — content-aware widths (floor at label width + sort-arrow reserve, col-0 leading space, baked 1-cell gap, slack to flex columns, flex-first shrink) — the #392 algorithm, generalized
- `RenderRow` — plain gapped/truncated/scrolled row text (the view applies styling)
- rune-aware `TruncateRunes` / `ScrollWindow` — **fixes the byte-slicing bug** in the old labels scroll for multibyte names/labels
- `ColumnDefs` helper to keep header labels in lockstep with the column set

`FilterableList` gains opt-in `Columns` + `columnScroll` fields and `ColWidths`/`RenderRow`/`ScrollLeft`/`ScrollRight`/`ResetColumnScroll`. **All new behavior is guarded by `Columns != nil`**, so the list views that don't opt in (stacks, contexts, networks, tasks) are byte-for-byte unchanged.

**Adoption:**
- **secrets / configs / nodes** declare their columns and delegate widths + base row text to the shared layer, keeping their own styling. The labels-only `>` scroll is replaced by the unified flex-column `…` scroll (NAME/ID/LABELS, and ID/HOSTNAME/LABELS for nodes). **Configs also gains the scroll-reset-on-cursor-move it was silently missing.**
- **services** migrated onto the shared layer; its bespoke `computeColWidths`/`distributeSlack`/`shrink`/`displayWidth`/`formatErrorWithScroll` are deleted (single source of truth). Its extras — conditional STACK column, error-row coloring, full-width highlight, expandable task sub-rows — stay view-side. The #392 alignment regression test now guards the shared engine.
- **"SECRET USED" / "CONFIG USED" → "USED"**, reclaiming ~7 columns for the flex columns (in the spirit of #394). Visible header text change.

Dead code removed: nodes `calcColumnWidths` + its unused `colWidths` map; the per-view byte-based `truncateWithEllipsis`/`formatLabelsWithScroll` copies.

## Tests
- New `layout_test.go`: width fit/shrink, gap + label-floor preserved when narrow, sort-arrow floor, header/row length, rune-aware truncate/scroll (incl. CJK).
- Per-view `columns_test.go` (secrets/configs/nodes/services): content-aware width (full name/hostname on a wide terminal), **header↔row width alignment** across widths/scopes/sort fields, left/right arrow shifts the window, reset-on-cursor-move, and the "USED" rename.
- `go test ./...`, `go vet`, `gofmt`, and `golangci-lint run ./ui/... ./views/...` all clean.

### DIFF
--- a/ui/components/filterable/list/filerablelist.go
+++ b/ui/components/filterable/list/filerablelist.go
@@ -39,9 +39,16 @@ type FilterableList[T any] struct {
 	// Nil means no footer is rendered by RenderFooter/RenderFramedView.
 	Footer *FooterConfig
 
-	outerWidth  int
-	outerHeight int
-	colWidth    int
+	// Columns, when non-nil, enables the shared content-aware table layout:
+	// ColWidths sizes columns to content, RenderRow builds gapped/scrolled rows,
+	// and ScrollLeft/Right drive a unified horizontal scroll. Leave nil to keep
+	// the legacy equal/Pct width behavior unchanged.
+	Columns []Column[T]
+
+	outerWidth   int
+	outerHeight  int
+	colWidth     int
+	columnScroll int // unified horizontal scroll offset for flex columns (shared layout)
 }
 
 // SetOuterSize stores fallback dimensions from tea.WindowSizeMsg.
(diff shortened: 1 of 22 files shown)


mosonyi:
LGTM!

PR #396, opened 2026-06-07
feat(volumes): read-only Docker volumes view (#260)

### LINKED ISSUE #260: volumes
docker volume list

```name, stack, driver, mount point, created, host```

docker volume ls is per-node, so we need the agents in place.

eldara-cruncher:
Closes #260. Lays the CE groundwork for the BE follow-up Eldara-Tech/swarmcli-be#204.

## What
A read-only `:volume` view listing Docker volumes with the requested columns: **NAME, STACK, DRIVER, MOUNT POINT, CREATED, HOST**. Supports column sorting (`N`/`S`/`D`/`C`/`H`), `/` filtering, and `i`/`enter` JSON inspect. Mirrors the `networks` view, minus all mutation.

## Why read-only / connected-node-only
`docker volume ls` is **per-node** and the manager API cannot read another node's volume state, so CE lists the **connected node only** with fully correct data. The `HOST` column is populated with the connected node's hostname. Cross-node aggregation, creation, and file browse/rename are owned by BE #204 (which uses the licensed agent).

## BE integration seams (no CE changes needed later)
- **Data:** new `docker.VolumeOps` interface + `Deps.Volumes`, returning an edition-agnostic `VolumeInfo`. BE swaps in an agent-backed all-nodes implementation.
- **Messaging:** a `volumes-all-nodes` feature flag gates a footer hint that all-node listing is a Business Edition feature (reuses `view.BELandingURL`). BE calls `features.Enable(...)` to suppress it.

Both are generic extension points — no pro internals in OSS.

## Changes
- `docker/volume.go`, `docker/volume_ops.go`, `docker/ops.go` — ops layer + `VolumeInfo`
- `views/volumes/` — the view (read-only)
- `commands/command/docker/volume/ls.go` + autoload wiring; `view.NameVolumes` top-level registration
- Unit tests (mapping, model behavior, column alignment, footer gating) + an integration test that creates a volume on the connected node and verifies the list/inspect path

## Verification
`go build -tags=integration ./...`, `go vet ./...`, `go test ./...`, and `golangci-lint run ./... --build-tags=integration` all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ TEST_LOG=1 ./test-setup/testenv.sh test
 | `:node`  | Navigate to Nodes       |
 | `:config`  | Navigate to Config    |
 | `:secret`  | Navigate to Secret    |
+| `:volume`  | Navigate to Volumes   |
 | `:node`  | Navigate to Nodes       |
 | `l`      | View Logs               |
 | `s`      | Scale Service           |
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ TEST_LOG=1 ./test-setup/testenv.sh test
 | `:node`  | Navigate to Nodes       |
 | `:config`  | Navigate to Config    |
 | `:secret`  | Navigate to Secret    |
+| `:network` | Navigate to Networks  |
 | `:volume`  | Navigate to Volumes   |
 | `:node`  | Navigate to Nodes       |
 | `l`      | View Logs               |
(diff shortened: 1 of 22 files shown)


clangenb:
@mosonyi I tested it and it works. For now, it only shows the local volumes in the community edition. The reason is that we will have quite some unnecessary code in CE for listing volumes on other nodes that is not necessary in BE.

I also thought that being able to see volumes on other nodes in BE only is an even stronger compelling incentive to go BE. Re-open the issue if you disagree.

PR #399, opened 2026-06-07
feat(ports): add network healthcheck between ports

Arka-cell:
### About Issue Eldara-Tech/swarmcli-be-releases#14 

To test; use the filter `/` then write `ports` or `port`. 

Did smoke sanity tests and tests seem good. TCP and UDP dialing was done with Claude but I'll still do research and try to separate it and send those packets to a VM where I'll block the port using a Firewall. Is it a good approach?

Also, for testing I used a Swarm with containers using [docker:29-dind](https://hub.docker.com/layers/library/docker/29-dind/images/sha256-df7efe7183e533112a8236c5fddb1edf12c96fb2feba3b6b693a285bd7bb993b) and I still haven't run another service in the Swarm, we'll look into it tomorrow.

### DIFF
--- a/commands/command/ports.go
+++ b/commands/command/ports.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package command
+
+import (
+	"swarmcli/args"
+	"swarmcli/registry"
+	portsview "swarmcli/views/ports"
+	"swarmcli/views/view"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type Ports struct{}
+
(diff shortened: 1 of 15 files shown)


clangenb:
I have been checking the concept of this, and I realized that this feature cannot be meaningfully implemented in the community edition. We need our agents that we install on each node to reliably get a health-check.

What you implement is a client>node healthcheck, but we need a node<>node healtcheck to get a cluster health state.

Thanks a lot for your effort, but I guess this feature cannot be implemented in the open source code. 🙏 


Apart from this, yes dind integration tests to test connectivity end to end would be a good approach.

clangenb:
I will close this one, as I do not think that you can do it in the Open Source version

PR #401, opened 2026-06-08
fix(logs): show starting-container logs in hide-stopped filter (#388)

### LINKED ISSUE #388: [LOG] Don't show stopped container logs

eldara-cruncher:
## Problem

Closes #388 (follow-up to [comment 4644247107](https://github.com/Eldara-Tech/swarmcli/issues/388#issuecomment-4644247107)).

PR #390 added the hide-stopped log filter but it had a regression: **starting containers' logs were not shown.**

The filter was an *allowlist that fails closed* — a line was shown only if its task's `DesiredState == Running` (`runningTaskIDs` + `lineVisible`). Any task not provably desired-running was hidden, so a starting container's logs disappeared because:

1. The task is not in the 3s-TTL snapshot yet (scale-up / crash-restart / deploy).
2. It is transiently in a non-`Running` desired state during start-first rolling updates.

## Fix

Invert to a *denylist that fails open*:

- `model.go`: `runningTaskIDs` → `stoppedTaskIDs`, collecting tasks whose **current `Status.State` is terminal** (complete/shutdown/failed/rejected/orphaned/remove) via new `isTerminalTaskState` helper. `lineVisible` now hides a line only if its task ID is in that stopped set.
- `update.go`: both call sites (`highlightContent`, `buildContent`) updated.
- Fail-open preserved: nil snapshot hides nothing; a task ID absent from the snapshot (starting/new) stays visible.

Net behavior: genuinely stopped/old containers stay hidden (original #388 still fixed); running, starting, and not-yet-tracked containers show.

## Tests

- Fixture switched from `DesiredState` to `Status.State`, added a starting task.
- New `TestBuildContent_ShowsStartingAndUnknownTasks` regression guard for the comment.
- `go test ./views/logs/` passes.

### DIFF
--- a/views/logs/logs_test.go
+++ b/views/logs/logs_test.go
@@ -691,14 +691,15 @@ func TestShouldFallbackToRawFromStdCopy(t *testing.T) {
 
 // --- hide-stopped (issue #388) tests ---
 
-// hideStoppedModel returns a model whose snapshot has one running and one
-// stopped task for the model's service (svc-123).
+// hideStoppedModel returns a model whose snapshot has one running, one starting
+// and one stopped (terminal) task for the model's service (svc-123).
 func hideStoppedModel() *Model {
 	m := testModel()
 	m.deps.Snapshot = &mockSnapshotOps{snap: &docker.SwarmSnapshot{
 		Tasks: []swarm.Task{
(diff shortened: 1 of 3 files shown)


mosonyi:
LGTM

PR #402, opened 2026-06-08
feat(volumes): flexible content-aware columns like the other list views (#400)

### LINKED ISSUE #400: [Volumes] flexible column with like the other views
We have recently improved the UI for secrets, config etc. We should do the same for the volumes

eldara-cruncher:
Closes #400.

## What

Migrates the **volumes** view onto the shared content-aware column layout (`ui/components/filterable/list/layout.go`) that **secrets, configs, nodes, services** adopted in #394/#395. Volumes was the last list view still on the legacy engine.

Before, volumes used:
- `volumeColWidths` — fixed percentage-weighted widths (not content-aware)
- `truncateWithEllipsis` — byte-based slicing (splits multibyte runes)
- a hand-rolled header + row formatter kept in sync manually
- no horizontal scrolling of truncated cells

Now it gets, for free and consistently with its siblings: content-sized columns (no dead space on wide terminals, no premature truncation on narrow ones), a guaranteed inter-column gap, one column set driving both header and rows (can't drift), and unified rune-aware left/right scrolling of truncated flex cells.

**Flex columns:** NAME, MOUNT POINT, HOST (the unbounded-length ones). STACK, DRIVER, CREATED size to content.

## Changes

- **`model.go`** — declare `Columns` via `buildColumns()`; header switches to `ColumnDefs(cols)` + a `SortIndicator`. Drops the bespoke `ColumnDef` literal and `ColWidthsFunc`.
- **`update.go`** — delete `volumeColWidths` and `truncateWithEllipsis`; add `buildColumns`, `sortColumnIndex` (maps sort field → column index; **not 1:1**, since MOUNT POINT is unsortable so CREATED/HOST sit one column past their ordinal), a one-line `RenderRow` delegation, and `left`/`right` scroll keys with scroll-reset on cursor move.
- **`view.go`** — `FrameHeader` delegates to the shared `RenderHeader`; removes the hand-rolled `renderVolumesHeader` and the now-unused `sorting` import.
- **tests** — replace the `volumeColWidths`/`renderVolumesHeader` tests with shared-API equivalents; add content-aware + scroll regression tests mirroring the nodes view.

## Note (non-breaking)

The inter-column gap narrows from 2 spaces to 1 (`ColGap=1`) — a deliberate, intended consequence of adopting the shared layer, identical to what the sibling views did in #395. Purely visual; no API or data-model change.

## Verification

- `go build -v -o swarmcli .` — clean
- `go test ./views/... ./ui/...` — all pass (incl. new regression tests)
- `golangci-lint run ./views/volumes/...` — 0 issues

### DIFF
--- a/views/volumes/columns_test.go
+++ b/views/volumes/columns_test.go
@@ -10,29 +10,62 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestColWidths_SumToWidth(t *testing.T) {
-	for _, w := range []int{60, 80, 100, 120, 200} {
-		widths := (&Model{}).volumeColWidths(w)
-		require.Len(t, widths, 6)
-		sum := 0
-		for _, cw := range widths {
-			require.Greater(t, cw, 0)
-			sum += cw
+// longVolName is wide enough to overflow the NAME column on a narrow terminal
+// and to verify full visibility on a wide one.
+const longVolName = "volume_with_a_fairly_long_descriptive_name"
+
+// longVolume returns the loaded item whose name overflows, for scroll tests.
+func longVolume(m *Model) volumeItem {
+	for _, v := range m.volumesList.Filtered {
+		if v.Name == longVolName {
+			return v
 		}
-		// columns + 5 two-space separators reconstruct the effective width.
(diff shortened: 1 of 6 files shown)


mosonyi:
LGTM

PR #408, opened 2026-06-11
feat: list all services with :service / :svc (#377)

### LINKED ISSUE #377: list services
It is possible to show all the services not just the services in the stack.

:service (alias: :svc)

docker service ls

eldara-cruncher:
## What

Adds a top-level command `:service` (alias `:svc`) that lists **all** swarm services across the cluster — the equivalent of `docker service ls` — instead of only the services scoped to a single stack.

Closes #377.

## Why it's small

The services view already had a dormant `AllFilter` mode (with a conditional STACK column shown only in that mode), and the factory already defaults to it for an empty payload. Two gaps prevented it from working:

1. **No command opened it** — the view was only reachable by drilling into a stack.
2. **The data layer returned empty** — the `AllFilter` path called `LoadStackServices("")`, which matches nothing because every service's stack is normalised to a real label or `"-"`, never `""`.

This PR activates the existing UI rather than adding a new view.

## Changes

- `docker/service.go` — new `LoadAllServices()` (snapshot iteration, no stack filter).
- `docker/service_ops.go` — `LoadAllServices()` on the `ServiceOps` interface + default impl.
- `views/services/service.go` — `AllFilter` case now calls `LoadAllServices()`.
- `commands/command/docker/service/ls.go` — `:service` command + `:svc` alias; registered via `commands/autoload.go`.

## Tests

- Command unit test (name, navigation message, `svc`→`service` alias).
- Rewrote `TestLoadServicesForView_AllFilter` to assert routing to `LoadAllServices`.
- New integration test asserting `LoadAllServices()` lists every `demo_*` service.
- `go build`, `go test ./...`, and `golangci-lint run --build-tags=integration` all pass locally.

### DIFF
--- a/commands/autoload.go
+++ b/commands/autoload.go
@@ -10,5 +10,6 @@ import (
 	_ "swarmcli/commands/command/docker/network"
 	_ "swarmcli/commands/command/docker/node"
 	_ "swarmcli/commands/command/docker/secret"
+	_ "swarmcli/commands/command/docker/service"
 	_ "swarmcli/commands/command/docker/volume"
 )
--- a/commands/command/docker/service/ls.go
+++ b/commands/command/docker/service/ls.go
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package service
+
+import (
+	"swarmcli/args"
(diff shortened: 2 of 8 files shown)


mosonyi:
LGTM

PR #409, opened 2026-06-11
feat(view): FeatureLockedCmdFn seam for rich gated-feature prompts

eldara-cruncher:
## What

Adds a generic extension seam `view.FeatureLockedCmdFn` so an extension (swarmcli-be) can show a richer prompt when a license-gated feature is requested without entitlement, instead of the flat `BEUnavailableErr` text shown today.

- `view.FeatureLockedCmdFn func(featureName string) tea.Cmd` + `view.FeatureLockedCmd(...)` helper, mirroring the adjacent `BEUnavailableFormatFn`.
- All gated-action denial sites prefer the seam and **fall back to the existing local error dialog when it is unset**, so pure-OSS behaviour is unchanged:
  - `views/services/update.go` — shell (`x`), active-forwards (`w`), port-forward (`W`)
  - `views/secrets/update.go` — reveal-secret
  - `views/volumes/update.go` — `dispatchAction` (all volume actions)

## Why

Enables swarmcli-be#174: replace the flat "Shell requires a Business Edition license" error with the same inline license dialog (cluster ID, where-to-get-a-license URL, paste-key input) the `:license` `<s>` flow uses. The BE side (which owns the dialog) lands in a companion PR.

## Test

- `go test ./views/...` — adds `TestFeatureLockedCmd_*` in `views/view/edition_test.go`.
- `go vet` + `golangci-lint run` clean on the changed packages.

Refs Eldara-Tech/swarmcli-be#174

### DIFF
--- a/views/secrets/update.go
+++ b/views/secrets/update.go
@@ -382,6 +382,9 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 			}
 			action, ok := view.GetAction("reveal-secret")
 			if !ok {
+				if cmd := view.FeatureLockedCmd("Reveal Secret"); cmd != nil {
+					return cmd
+				}
 				m.err = view.BEUnavailableErr("Reveal Secret")
 				m.errorDialogActive = true
 				return nil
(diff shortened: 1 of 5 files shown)


clangenb:
I merge this on as it is only the seam for BE's dialog. Refinement will happen there.

PR #410, opened 2026-06-17
fix(contexts): allow switching to a locked swarm + in-app unlock

### LINKED ISSUE #421: Swarm is locked — cannot switch context
When the Swarm is locked, SwarmCLI cannot switch context. (See Docker's [swarm manager locking](https://docs.docker.com/engine/swarm/swarm_manager_locking/).) The command-line context switch still works.

1. The default (local) context is locked.
2. If the default is the current context, the stack isn't visible — expected (`docker stack ls` → "Swarm is encrypted and needs to be unlocked").
3. Switching context inside SwarmCLI fails, while `docker context use` works on the command line.

_(Original report included a screenshot showing SwarmCLI unable to switch away from the locked context.)_

eldara-cruncher:
## Problem
Resolves #421 ("Swarm is locked"; migrated to CE from Eldara-Tech/swarmcli-be#177). When a context's Swarm is locked (autolock), swarmcli refused to switch into it, even though `docker context use` works from the CLI.

A locked swarm is **reachable** — `/_ping` and `/info` succeed and `docker info` reports `LocalNodeState == "locked"` — but every store-backed call (`node ls`, `service ls`, `stack ls`, …) fails with *"Swarm is encrypted and needs to be unlocked"*. Two places treated that as fatal and reverted the switch:

- `ValidateContext` rejected any state `!= active`.
- `RefreshSnapshot` returned an error on `NodeList`; `app` then reverted to the previous context.

The fix lives in CE; BE consumes it via `replace swarmcli => ../swarmcli` and doesn't override these ops, so it flows to BE on rebuild.

## Changes
- `docker/swarm_lock.go` (new): `IsSwarmLockedErr`, `isUsableSwarmState`, `UnlockSwarm` (wraps `cli.SwarmUnlock`).
- `ValidateContext` accepts the locked state and tolerates a locked `Info` error → the switch succeeds (CLI parity).
- `SwarmSnapshot.Locked` flag; `RefreshSnapshot` returns an empty flagged snapshot instead of erroring → no revert.
- New `views/unlockdialog/` (password-masked text input) + `:unlock` command → calls `SwarmUnlock`, then invalidates and reloads.
- App shows a **neutral info** notice when landing on a locked swarm, explaining `:unlock` / `docker swarm unlock`.
- Unified startup / context-switch / post-unlock snapshot loading through one `snapshotLoadedMsg` path (removed the duplicate `loadSnapshotAndNavigateToStacksCmd`).

## Testing
- `go build`/`go vet` clean; `go test ./...` passes (new `docker/swarm_lock_test.go`, `views/unlockdialog/unlockdialog_test.go`).
- `golangci-lint run ./... --build-tags=integration` → 0 issues.
- `swarmcli-be` builds against the patched CE.
- The locked branches of `ValidateContext`/`RefreshSnapshot` need a live daemon — verify manually: autolock a test swarm, switch in (lands on stacks with the info notice, no revert), `:unlock` with the key (resources reload), wrong key → clear error + retry.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -90,3 +90,32 @@ jobs:
         with:
           name: test-report
           path: /tmp/test-report.xml
+
+  # The locked-swarm test autolocks and restarts its own throwaway dind daemon,
+  # so it must stay isolated from the shared multi-node swarm above. It is gated
+  # behind the dedicated `integration_locked` build tag and needs no compose
+  # bring-up, so it runs as its own job for a clean, independent signal.
+  locked-swarm:
+    needs: [changes]
+    if: needs.changes.outputs.integration == 'true'
+    name: Run locked-swarm integration test
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Verify Docker & Go
+        run: |
+          docker version
+          go version
+
+      - name: Run locked-swarm integration test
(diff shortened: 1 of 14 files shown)


clangenb:
@mosonyi ready to review. It runs the lock and unlock even from the integration tests.

mosonyi (approved):
Perfect!

PR #412, opened 2026-06-18
fix(logs): align node-selector overlay over special-char log lines

### LINKED ISSUE #369: Node selector at logs view
If a log is tailing and I want to choose a specific node only, the small overlay  is rendered faulty when a special character exists in the log

Image

eldara-cruncher:
## Problem

Fixes #369. While tailing logs, opening the **Select Node to Filter** overlay rendered a broken box — zig-zag borders with background log text bleeding through — but only when a log line behind the overlay contained a "special character" (wide CJK/emoji or an embedded escape sequence).

## Root cause

`ui.OverlayCentered` positioned the dialog using `lipgloss.Width` (grapheme/cell-width and full-ANSI aware) but *sliced* the background lines with the hand-rolled `TruncateANSI`/`TruncateANSIAfter`, which counted every rune as 1 cell and only recognised SGR escapes ending in `m`. On lines with wide chars or non-SGR escapes the two width models disagreed, so each dialog row spliced at a different column.

## Fix

- Reimplement `TruncateANSI`/`TruncateANSIAfter` over `charmbracelet/x/ansi` (`ansi.Truncate`/`ansi.TruncateLeft`) — the same engine `lipgloss.Width` uses — so slicing and measuring share one cell-width model. Signatures unchanged; no call-site changes. This also fixes the logs horizontal-scroll path that shares these helpers.
- Pad the overlay's left part to the start column so the box stays aligned when a wide grapheme straddles the cut.
- Promote `x/ansi` to a direct dependency (`go mod tidy`).
- Add the previously-missing unit tests for both helpers plus a #369 regression test for `OverlayCentered`.

## Verification

- `go build`, `go test ./...`, `golangci-lint run ./...` all pass.
- The regression test fails against the pre-fix code (dialog border lands at column 10 on some rows, 15 on others — the exact bug) and passes after the fix.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/charmbracelet/bubbles v1.0.0
 	github.com/charmbracelet/bubbletea v1.3.10
 	github.com/charmbracelet/lipgloss v1.1.0
+	github.com/charmbracelet/x/ansi v0.11.6
 	github.com/docker/docker v28.5.2+incompatible
 	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/muesli/reflow v0.3.0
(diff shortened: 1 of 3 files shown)


mosonyi:
LGTM

mosonyi (approved):

PR #415, opened 2026-06-21
feat(charts): Helm-like package manager for Docker Swarm (Phase 1 + early Phase 2)

### LINKED ISSUE #413: charts
# SwarmCLI Charts

SwarmCLI Charts is a package manager for Docker Swarm inspired by Helm.

Charts provide a standardized way to package, distribute, configure, install, upgrade, and manage applications running on Docker Swarm.

Official chart repository:

https://github.com/Eldara-Tech/swarmcli-charts

---

# Concepts

## Chart

A chart is a versioned application package containing:

* Docker Stack templates
* Default configuration values
* Metadata
* Documentation
* Dependencies

Example charts:

* traefik
* airflow
* superset
* harbor
* minio
* keycloak

## Release

A release is a deployed instance of a chart.

Examples:

```bash
swarmcli charts install traefik-prod eldara/traefik
swarmcli charts install traefik-dev eldara/traefik
```

Both releases use the same chart but have different configurations.

---

# Repository Management

## Add Repository

```bash
swarmcli charts repo add eldara https://github.com/Eldara-Tech/swarmcli-charts
```

## List Repositories

```bash
swarmcli charts repo list
```

Example:

```text
NAME     URL
eldara   https://github.com/Eldara-Tech/swarmcli-charts
```

## Update Repository Metadata

```bash
swarmcli charts repo update
```

Downloads the latest chart index from all configured repositories.

## Remove Repository

```bash
swarmcli charts repo remove eldara
```

---

# Discovering Charts

## Search Charts

```bash
swarmcli charts search traefik
```

Example:

```text
NAME              VERSION
eldara/traefik    3.0.1
eldara/airflow    2.10.0
eldara/superset   5.0.0
```

## Search All Charts

```bash
swarmcli charts search
```

## Show Chart Information

```bash
swarmcli charts show chart eldara/traefik
```

Displays:

* Name
* Description
* Maintainer
* Version
* App Version
* Dependencies
* README

---

# Working with Values

## Show Default Values

```bash
swarmcli charts show values eldara/traefik
```

Displays the chart's default `values.yaml`.

## Download Values

```bash
swarmcli charts show values eldara/traefik > values.yaml
```

## Show Values Schema

```bash
swarmcli charts show schema eldara/traefik
```

Displays all supported configuration options.

---

# Template Rendering

## Render Templates

```bash
swarmcli charts template my-traefik eldara/traefik
```

Renders templates locally without deploying.

## Render Using Custom Values

```bash
swarmcli charts template my-traefik eldara/traefik \
  --values values.yaml
```

## Render with Inline Overrides

```bash
swarmcli charts template my-traefik eldara/traefik \
  --set replicas=3
```

Useful for:

* CI/CD validation
* Troubleshooting
* Reviewing generated stacks
* GitOps workflows

---

# Installing Charts

## Install

```bash
swarmcli charts install my-traefik eldara/traefik
```

## Install with Custom Values

```bash
swarmcli charts install my-traefik eldara/traefik \
  --values values.yaml
```

## Install with Inline Overrides

```bash
swarmcli charts install my-traefik eldara/traefik \
  --set replicas=3 \
  --set image.tag=v3.5.0
```

## Dry Run Installation

```bash
swarmcli charts install my-traefik eldara/traefik \
  --dry-run
```

No resources are deployed.

---

# Upgrading Releases

## Upgrade

```bash
swarmcli charts upgrade my-traefik eldara/traefik
```

## Upgrade with Values

```bash
swarmcli charts upgrade my-traefik eldara/traefik \
  --values values.yaml
```

## Upgrade or Install

```bash
swarmcli charts upgrade \
  --install \
  my-traefik \
  eldara/traefik
```

Installs the release if it does not exist.

## Preview Upgrade Changes

```bash
swarmcli charts diff upgrade \
  my-traefik \
  eldara/traefik
```

Shows changes before deployment.

---

# Release Management

## List Releases

```bash
swarmcli charts list
```

Example:

```text
NAME           CHART
my-traefik     traefik
airflow-prod   airflow
```

## Show Release Status

```bash
swarmcli charts status my-traefik
```

Displays:

* Chart version
* Release version
* Service health
* Deployment state

## Show Release Values

```bash
swarmcli charts get values my-traefik
```

Displays values used during deployment.

## Show Generated Manifest

```bash
swarmcli charts get manifest my-traefik
```

Displays the rendered Docker Stack.

## Show Release History

```bash
swarmcli charts history my-traefik
```

Example:

```text
REVISION  CHART VERSION
1         3.0.0
2         3.0.1
3         3.1.0
```

---

# Rollbacks

## Roll Back Release

```bash
swarmcli charts rollback my-traefik 2
```

Re-deploys revision 2.

---

# Uninstalling

## Remove Release

```bash
swarmcli charts uninstall my-traefik
```

Removes:

* Docker services
* Configs
* Secrets
* Networks

Volumes remain by default.

## Remove Including Volumes

```bash
swarmcli charts uninstall my-traefik \
  --purge-volumes
```

---

# Dependency Management

## Update Dependencies

```bash
swarmcli charts dependency update ./mychart
```

Downloads required chart dependencies.

## Build Dependencies

```bash
swarmcli charts dependency build ./mychart
```

---

# Chart Development

## Create Chart

```bash
swarmcli charts create mychart
```

Creates:

```text
mychart/
├── Chart.yaml
├── values.yaml
├── README.md
├── templates/
│   ├── stack.yaml
│   ├── configs.yaml
│   ├── secrets.yaml
│   └── volumes.yaml
└── charts/
```

## Validate Chart

```bash
swarmcli charts lint ./mychart
```

Checks:

* Metadata
* Template syntax
* Values
* Dependencies

# Common Flags

## Values File

```bash
-f values.yaml
--values values.yaml
```

## Override Values

```bash
--set key=value
```

## Dry Run

```bash
--dry-run
```

## Wait for Deployment

```bash
--wait
```

## Timeout

```bash
--timeout 10m
```

## Debug Output

```bash
--debug
```

---

# Typical Workflow

```bash
swarmcli charts repo add eldara https://github.com/Eldara-Tech/swarmcli-charts

swarmcli charts repo update

swarmcli charts search traefik

swarmcli charts show values eldara/traefik > values.yaml

swarmcli charts template my-traefik eldara/traefik \
  -f values.yaml

swarmcli charts install my-traefik eldara/traefik \
  -f values.yaml

swarmcli charts status my-traefik

swarmcli charts upgrade my-traefik eldara/traefik \
  -f values.yaml

swarmcli charts history my-traefik

swarmcli charts rollback my-traefik 2
```

# Release Storage

SwarmCLI Charts stores release metadata and revision history in Docker Swarm Configs.

This approach provides:

* High Availability through the Swarm Raft database
* Automatic replication across manager nodes
* No external database required
* Native Docker Swarm integration
* Fast rollback operations

## Config Naming

Each release revision is stored as a separate Docker Config.

Example:

```text
swarmcli.release.traefik-prod.v1
swarmcli.release.traefik-prod.v2
swarmcli.release.traefik-prod.v3
```

## Config Labels

Every release Config contains metadata labels.

Example:

```text
com.swarmcli.type=release
com.swarmcli.release=traefik-prod
com.swarmcli.chart=traefik
com.swarmcli.chart.version=3.0.1
com.swarmcli.revision=3
com.swarmcli.status=deployed
com.swarmcli.created=2026-06-21T14:30:00Z
```

## Config Content

The Config payload contains the complete release definition.

Example:

```yaml
release: traefik-prod
revision: 3

chart:
  name: traefik
  version: 3.0.1

values:
  replicas: 3
  image:
    tag: v3.5.0

manifest: |
  version: "3.9"

  services:
    traefik:
      image: traefik:v3.5.0
      deploy:
        replicas: 3
```

## Listing Stored Revisions

Docker administrators can inspect release history directly:

```bash
docker config ls \
  --filter label=com.swarmcli.release=traefik-prod
```

Example:

```text
ID         NAME                               REVISION
abcd1234   swarmcli.release.traefik-prod.v1   1
efgh5678   swarmcli.release.traefik-prod.v2   2
ijkl9012   swarmcli.release.traefik-prod.v3   3
```

## Release History

SwarmCLI reads release revisions from Docker Config labels.

```bash
swarmcli charts history traefik-prod
```

Output:

```text
REVISION  STATUS      CHART VERSION
1         superseded  3.0.0
2         superseded  3.0.1
3         deployed    3.1.0
```

## Rollback

Rollback creates a new revision based on a previous revision.

```bash
swarmcli charts rollback traefik-prod 1
```

Result:

```text
REVISION  STATUS
1         superseded
2         superseded
3         superseded
4         deployed
```

Revision 4 contains the contents of revision 1.

Like Helm, release history is append-only and fully auditable.

## Cleanup

By default, SwarmCLI keeps all revisions.

Administrators can limit retained history:

```bash
swarmcli charts install traefik-prod eldara/traefik \
  --history-max 20
```

Older revisions are automatically removed once the limit is reached.

## Disaster Recovery

Because release history is stored in Docker Configs, it is automatically replicated through the Swarm manager quorum.

No additional backup database is required.

Release metadata survives:

* Manager node failures
* Leader elections
* Swarm failovers

as long as manager quorum remains available.

eldara-cruncher:
## What

**SwarmCLI Charts** (#413): a Helm-inspired package manager for Docker Swarm. Charts package Docker Stack (Compose) templates + values + metadata; installing a chart produces a **release** — a Docker stack whose append-only revision history is stored in Docker Configs.

> **Scope note.** This PR was opened as "Phase 1 (MVP)" but, as built, it also lands the most useful early **Phase 2** release-lifecycle commands and the **TUI guards** that protect chart-managed resources. The list below reflects what actually ships; the original Phase-2 follow-up bullet has moved up accordingly. This expansion is intentional and confirmed with the maintainer.

**Subcommands this PR ships:**
- Repo/discovery: `repo add/list/update/remove`, `search`, `show chart|values|schema`, `template`.
- Lifecycle: `install`, `uninstall`, `list`, `status`.
- Release management (early Phase 2): `upgrade [--install]`, `rollback`, `history`, `diff`, `get values|manifest`.

**TUI guards (this PR):** chart-managed stacks and configs are protected from destructive TUI actions; a popup explains why editing a chart-release config is blocked; gzip-compressed release-record configs render as text in the config viewer.

## Key design decisions (confirmed with maintainer)

- **CLI-first execution.** swarmcli was TUI-only (`main.go` ignored `os.Args`). The issue is written entirely as scriptable CLI (stdout, `> values.yaml`, `--dry-run` for CI, GitOps). So `main.go` now dispatches to a new `cli/` package when args are present; a bare `swarmcli` still launches the TUI unchanged.
- **Repo format:** Helm-style HTTP `index.yaml` + `.tgz` tarballs.
- **Templating:** Go `text/template` + Sprig over `.Values`/`.Release`/`.Chart`, with optional `values.schema.json` validation. Each `templates/*.yaml` is deep-merged into one Compose document (Compose is a single document, unlike Helm's concatenated manifests).
- **Release storage:** one immutable, gzipped Docker Config per revision (`swarmcli.release.<rel>.v<N>`) labeled `com.swarmcli.*`; `release = stack namespace`. Statuses (`superseded`) are *derived* from the append-only log rather than mutating immutable Configs. HA via Swarm Raft, no external DB.

## Layers

- `cli/` — non-interactive front-end (arg parsing, stdout tables, exit codes).
- `charts/` — pure-Go engine (`types`, `values`, `render`, `chart`, `repo`, `release`, `backend_docker`), reusable by a future TUI browser view.
- `docker/` — two small symmetric helpers added: `RemoveStackCLI` (full `docker stack rm` teardown) and `RemoveVolume` (for `--purge-volumes`).

## Notes / limitations (documented in `charts/README.md`)

- `install --dry-run` still contacts Docker to compute the next revision; use `charts template` for fully offline rendering (mirrors Helm).
- `--purge-volumes` is connected-node only (CE volume scope).
- **Security hardening (in this PR).** Untrusted-chart exposure is reduced at the engine boundary: the host-reaching Sprig helpers (`env`/`expandenv`/`getHostByName`) are denied so a template cannot exfiltrate host env/DNS into the rendered manifest; the chart tarball download is restricted to `http(s)` with a redirect guard (SSRF / scheme-downgrade) and bounded by per-entry, entry-count and cumulative-size caps. The rendered manifest still lives in a Docker Config readable by anyone with Docker access, so templates must reference Docker secrets as separate objects, not inline secret material; a broader redaction pass remains planned before wide rollout.
- **Robustness.** Install is atomic — auto-created external networks are rolled back if the deploy fails; release recording retries on a concurrent revision-number collision; uninstall continues cleanup on partial failure and `stack rm` of an absent stack is idempotent.
- Uses `docker stack deploy` under the hood → Compose-on-Swarm subset only; `docker` CLI must be on `PATH`.

## Testing

- `go build ./...`, `go test ./...`, `golangci-lint run ./... --build-tags=integration` all pass.
- Unit tests cover values precedence, rendering/gating, schema validation, archive round-trip, repo index/search/resolve/pull, and the full release lifecycle via a fake backend — plus error paths (archive/render/values/repo failures) and the hardening above (env denial, non-http(s) rejection, archive caps, revision-collision retry, network rollback, uninstall continue-on-failure).
- `integration-tests/charts/` exercises real install→list→status→uninstall under `//go:build integration` (needs the DinD swarm env).

## Follow-ups

- Phase 2 remainder: `--history-max` pruning UX polish, richer `diff`/`get` output.
- Phase 3: `create`, `lint`, `dependency` (subcharts); optional read-only `:charts` TUI browser.
- Seed `Eldara-Tech/swarmcli-charts` with real charts + published `index.yaml`.

Closes #413.


mosonyi:
If the index download fails, do not add repo.

```
vscode ➜ /workspaces/swarmcli (feat/413-charts-phase1) $ ./swarmcli charts repo add eldara https://github.com/Eldara-Tech/swarmcli-charts
Error: repository added but index download failed: update "eldara": fetch https://github.com/Eldara-Tech/swarmcli-charts/index.yaml: HTTP 404 Not Found
vscode ➜ /workspaces/swarmcli (feat/413-charts-phase1) $ ./swarmcli charts repo add eldara https://github.com/Eldara-Tech/swarmcli-charts
Error: repository "eldara" already exists
```

eldara-cruncher:
Fixed in 5003902.

The root cause was twofold:

**1. `charts repo add` wasn't atomic.** It saved the repo to `repos.json` *before* downloading the index, so a 404 left the repo half-added (`repository added but index download failed`) and a retry then hit `already exists`. It now downloads and validates the index first and only persists on success — a failed add leaves nothing behind and is immediately retryable.

**2. The 404 itself was a wrong URL.** `https://github.com/Eldara-Tech/swarmcli-charts` is the repo's HTML page, not the chart index — the index is served via GitHub Pages at `https://eldara-tech.github.io/swarmcli-charts`. The swarmcli-charts docs pointed at the wrong host (and used `chart` instead of `charts`); fixed in Eldara-Tech/swarmcli-charts#2.

To catch this class of mistake at the source, `repo add` now hints the Pages URL when handed a `github.com` URL:

```
$ swarmcli charts repo add eldara https://github.com/Eldara-Tech/swarmcli-charts
Error: index download failed, repository not added: fetch https://github.com/Eldara-Tech/swarmcli-charts/index.yaml: HTTP 404 Not Found
hint: github.com serves repository pages, not chart indexes; use the GitHub Pages URL https://eldara-tech.github.io/swarmcli-charts
```

Covered by unit tests `TestRepoStoreAddIndexFailureDoesNotPersist` and `TestGithubPagesHint`.

mosonyi:
I got an error message:
```
 $ ./swarmcli charts install whoami eldara/whoami --set ingress.host=whoami.test.com
Error: deploy failed: failed to deploy stack: exit status 1
Output: Since --detach=false was not specified, tasks will be created in the background.
In a future release, --detach=false will become the default.
network "traefik-public" is declared as external, but could not be found. You need to create a swarm-scoped network before the stack is deployed
```

In SwarmCLI — before running docker stack deploy, iterate over all networks: in the rendered stack file that have external: true and check existence via the Docker API (/networks endpoint filtered by name + scope=swarm). If any are missing, abort with a helpful message listing the docker network create commands needed.
Option B is more robust since it works for any chart without chart authors having to remember to declare requirements. Worth adding as a pre-flight check in the charts install command path.

mosonyi:
```
$ ./swarmcli charts install whoami eldara/whoami --set ingress.host=whoami.test.com
Error: deploy failed: failed to deploy stack: exit status 1
Output: Since --detach=false was not specified, tasks will be created in the background.
In a future release, --detach=false will become the default.
network "traefik-public" is declared as external, but could not be found. You need to create a swarm-scoped network before the stack is deployed
```
I expect not to install in this case, which is the case - OK

```
$ ./swarmcli charts install whoami eldara/whoami --set ingress.host=whoami.test.com
Error: release "whoami" already exists (revision 1); use upgrade
```

When deploy was not successful, the config should not be created.
```
 $ docker config ls --filter 'name=swarmcli.release.'
ID                          NAME                         CREATED          UPDATED
w2e6bxd7sh4k7zzkvbgq8flsj   swarmcli.release.whoami.v1   36 minutes ago   36 minutes ago
```

eldara-cruncher:
Both addressed in a658cec.

**1. Failed deploy no longer leaves an orphaned release (your `already exists` finding).**
`Engine.Install` was recording the `swarmcli.release.<name>.v1` Config even when the deploy failed, so a retry was blocked and the orphan config lingered. It now records **nothing** on deploy failure — a failed install leaves no config and is immediately retryable.

**2. Missing external networks (`traefik-public`) — auto-provisioned.**
Rather than only aborting, `charts install` now pre-flights the external networks a chart declares and **auto-creates** missing ones as attachable swarm overlays, then deploys:

- exists & swarm-scoped → used as-is
- absent → `docker network create --driver overlay --attachable <name>` (done for you)
- exists but **not** swarm-scoped, or create fails (e.g. not a swarm manager) → abort *before* deploy with the exact manual commands to run

> Note: this diverges slightly from your "abort + list the commands" suggestion — @clangenb opted to proactively create the network, keeping the manual-command message only as the fallback when auto-create isn't possible. Happy to switch to abort-only if you'd prefer the more conservative behavior.

The check is scoped to the charts install path (`docker.DeployStack`, shared with the stacks view, is untouched) and runs zero Docker calls for charts without external networks.

Coverage: unit tests for the manifest parser and all four pre-flight branches (present / auto-created / create-fails / non-swarm clash) plus the no-record-on-failure retry; and an integration test (`TestChartsExternalNetworkAutoCreate`) that installs a chart with a missing external network and asserts it's auto-created swarm-scoped and deploys against the live Swarm.

eldara-cruncher:
Added commit `b6a69bb` — `fix(charts): resolve chart versions with or without the "v" prefix`.

Context: a user hit `chart "whoami" version "v0.1.3" not found`. Chart versions are plain SemVer (`0.1.3`); the release **git tag** carries the `v` (`whoami/v0.1.3`), so people naturally type `--version v0.1.3`. `Resolve()` did a byte-exact match and rejected it. Now it matches SemVer-aware via the existing `compareVersions` helper (normalizes both sides, falls back to exact string equality for non-SemVer). Added `TestResolveVersionPrefixNormalization`.

The release-pipeline side of the same report (the published `.tgz` carried a stale `version: 0.1.0` internally) is fixed in swarmcli-charts#6.

🤖 via Claude Code

mosonyi:
```
$ ./swarmcli charts install whoami swarmcli-charts/whoami --version v0.1.6 --set ingress.host=whoami.test.com
Error: release "whoami" already exists (revision 1); use upgrade
```

But no upgrade function exists

mosonyi:
Do not allow to delete a config from tui, when it belongs to a chart install. Or at least ask if 'are you sure because it is swarmchart'...

mosonyi:
This command puts the version 3.9 to the end of the file? same with --dry-run

```
$ ./swarmcli charts template whoami swarmcli-charts/whoami --set ingress.host=whoami.test.com
networks:
  traefik-public:
    external: true
services:
  whoami:
    deploy:
      labels:
        - traefik.enable=true
        - traefik.docker.network=traefik-public
        - traefik.http.routers.whoami-http.rule=Host(`whoami.test.com`)
        - traefik.http.routers.whoami-http.entrypoints=web
        - traefik.http.routers.whoami-http.middlewares=redirect-to-https@file
        - traefik.http.routers.whoami-https.rule=Host(`whoami.test.com`)
        - traefik.http.routers.whoami-https.entrypoints=websecure
        - traefik.http.routers.whoami-https.tls=true
        - traefik.http.routers.whoami-https.tls.certresolver=letsencrypt
        - traefik.http.services.whoami.loadbalancer.server.port=80
      replicas: 2
      restart_policy:
        condition: on-failure
    image: ghcr.io/traefik/whoami:v1.11.0
    networks:
      - traefik-public
version: "3.9"
```

mosonyi:
If it is not really updated, say it is already up-to-date.

```
vscode ➜ /workspaces/swarmcli (feat/413-charts-phase1) $ ./swarmcli charts repo update
Updated 1 repository: swarmcli-charts
vscode ➜ /workspaces/swarmcli (feat/413-charts-phase1) $ ./swarmcli charts repo update
Updated 1 repository: swarmcli-charts
vscode ➜ /workspaces/swarmcli (feat/413-charts-phase1) $ 
```

eldara-cruncher:
Folded **Phase 2 (#416)** into this PR so install and upgrade ship together — per the thread, install-only is a dead-end (you immediately hit `already exists; use upgrade` with no `upgrade` command). Phase 2's single commit was cherry-picked onto the current phase-1 tip (linear, no force-push), commit `37de09c`.

**Rebase had one semantic conflict** in `charts/release.go`: Phase 2 refactored `Install` into a shared `deployAndRecord()` (reused by Upgrade/Rollback) that **recorded a failed revision on deploy failure** and **lacked `ensureExternalNetworks`** (it predated the atomic-install fix `a658cec`). Recording on failure would reintroduce the "already exists" orphan bug — a `Failed` revision becomes `currentRevision`, and Install blocks on `cur.Status != StatusUninstalled`. Resolved by keeping Phase 1's atomic behavior in the shared helper (no record on failure → retryable) and adding `ensureExternalNetworks`. No Phase 2 test asserted the failure-recording path, so nothing else changed.

`go build`, `vet`, `gofmt`, and `go test ./charts/ ./cli/` all green locally (incl. Phase 2's `upgrade_test.go`/`diff_test.go`). #416 closed in favor of this.

mosonyi:
At config view when I press enter on the charts config view I got this messy screen

image

eldara-cruncher:
@mosonyi addressed all three in `238d74f`:

1. **Don't delete a chart's config from the TUI without warning** — deleting a Docker config that backs a chart release (labeled `com.swarmcli.type=release`) now shows a confirmation that names the release and warns it may corrupt it (pointing at `charts uninstall`). You can still override explicitly, per your "or at least ask". Covered by `TestDeleteConfirmPrompt`.

2. **`version: "3.9"` at the end of the manifest** — that key is obsolete for `docker stack deploy` (the engine ignores it) and only showed up trailing because the merged manifest is re-encoded from a sorted map. The renderer now drops it, so `template`/`--dry-run` output is clean. Verified no top-level `version:` remains.

3. **`repo update` always said "Updated"** — it now compares the fetched index against the cache and prints `Already up-to-date: <repo>` when nothing changed (the served index is byte-stable between releases). Running it twice now reports up-to-date.

All with unit tests; `build`/`vet`/`gofmt`/`go test ./charts/ ./cli/ ./views/configs/` green locally.

mosonyi:
Config protection is ok. The same stack protection is needed.

eldara-cruncher:
@mosonyi (1/2) the **messy screen** is fixed in `d1bab96`: chart release records are stored gzip-compressed, so the config inspect/raw views were rendering binary. `ConfigWithDecodedData.DisplayData()` now transparently decompresses gzip payloads, so pressing enter on a chart's config shows the readable release record instead of garbage.

The **stack protection** (2/2) needs a design call — replying separately.

mosonyi:
Messy screen is fixed. With the <e> edit & rotate option, I could ruin the install, do not allow from tui.

eldara-cruncher:
@mosonyi both done in `b01d633`:

- **Stack protection** — deleting a stack that backs a chart release now pops the same kind of guard: it warns *"Stack X belongs to chart release Y — prefer `charts uninstall`"* before removing. Detection is by looking up the release config labeled with the stack name (works for existing releases too, no engine change), done off the UI thread.
- **`<e>` edit & rotate** — now blocked outright for chart release configs (rotating a gzipped release record would corrupt the install). It tells you to use `charts upgrade` instead.

Unit-tested; `build`/`vet`/`gofmt`/`go test ./views/stacks/ ./views/configs/ ./docker/` green.

mosonyi:
I can see, pressing <e> is doing nothing, I would prefer a popup to tell why it is not possible.

eldara-cruncher:
@mosonyi good catch — the old block used `tea.Printf`, which doesn't surface in the TUI (hence "nothing happens"). Fixed in `0db94b9`: pressing `<e>` on a chart release config now shows a dismissable **info popup** explaining it belongs to chart release X and to use `charts upgrade`. (Mode flags reset on dismiss so a later delete confirm still works normally.) Covered by `TestEditBlockedForChartConfig`.

mosonyi:
image

The y and n do nothing. ESC is ok. I would just remove the y/n section and add ESC - close

eldara-cruncher:
Good catch — fixed in a9a77c5.

Root cause: the edit-block popup was already shown in the dialog's dismiss-only `InfoMode` (so `y`/`n` were intentionally inert), but the configs view rendered it through `ui.RenderConfirmDialog`, a y/n-only renderer that ignores the mode — hence the misleading `<y> Yes • <n> No` footer.

Now dismiss-only Info/Error dialogs render via the `confirmdialog` component (footer reads **`<Enter/Esc> Close`**), matching how the stacks/services/nodes views already render. The y/n confirm path (e.g. delete) is unchanged. Added a regression test asserting the rendered popup shows "Close" and no y/n prompt.

PR #418, opened 2026-06-23
feat(charts): pre-flight external secrets/configs before deploy

eldara-cruncher:
Implements Eldara-Tech/swarmcli-charts#26.

## Why

External **networks** a stack needs are already auto-created on install (`ensureExternalNetworks`). External **secrets/configs** cannot be auto-created — their content is not part of a chart — so today a missing one surfaces only as a raw `docker stack deploy` failure with no guidance.

## What

A pre-flight check that, before any swarm state is mutated, verifies every secret/config the rendered manifest declares `external` already exists; otherwise it fails with the exact `docker secret/config create` commands to run (mirroring the existing network remediation message).

It runs **before** `ensureExternalNetworks`, so a missing prerequisite aborts the install **without leaving auto-created networks behind**.

- `networks.go`: extract `externalResourceNames(manifest, key)` — networks/secrets/configs share the compose `external` syntax (scalar `true` and long-form `{name: …}`); `externalNetworks` now delegates to it.
- `release.go`: `Backend.SecretNames` + `ensureExternalSecretsConfigs`, wired into `deployAndRecord` ahead of the network step.
- `backend_docker.go`: `SecretNames` via `docker.ListSecrets`; configs reuse the existing `ListConfigs`.

## Tests

Parser cases for secrets/configs (incl. long-form, `external:false`, unrelated complex keys), the pre-flight (present / missing-secret / missing-config / long-form / backend-error), and the **ordering guarantee** (missing secret aborts before any network is created). `go test -race ./charts/` green; gofmt + golangci-lint clean.

## Notes

- No chart in swarmcli-charts uses external secrets/configs yet — this is forward-looking infrastructure, unit-tested via the `Backend` fake (no daemon needed).
- Content is intentionally **not** chart-supplied (security); this is validation/messaging only.

### DIFF
--- a/charts/backend_docker.go
+++ b/charts/backend_docker.go
@@ -122,3 +122,15 @@ func (dockerBackend) RemoveOverlayNetwork(ctx context.Context, name string) erro
 	}
 	return nil // already gone
 }
+
+func (dockerBackend) SecretNames(ctx context.Context) (map[string]struct{}, error) {
+	secs, err := docker.ListSecrets(ctx)
+	if err != nil {
+		return nil, err
+	}
+	names := make(map[string]struct{}, len(secs))
+	for _, s := range secs {
+		names[s.Spec.Name] = struct{}{}
+	}
+	return names, nil
+}
--- a/charts/networks.go
+++ b/charts/networks.go
@@ -10,39 +10,55 @@ import (
 )
 
 // externalNetworks returns the real names of the networks a rendered stack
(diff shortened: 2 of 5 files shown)


mosonyi:
I think we can define the secrets and configs in the default values.yaml file and say in the requirements.yaml that those values or the secrets/config are compulsory to exist.

But  keep it simple, do how you suggested

PR #419, opened 2026-06-23
feat(charts): consume requirements.yaml pre-flight + uninstall network summary (#23, #28)

eldara-cruncher:
Implements Eldara-Tech/swarmcli-charts#23 (consume `requirements.yaml`) and Eldara-Tech/swarmcli-charts#28 (uninstall summary of auto-created networks). Companion charts PR: Eldara-Tech/swarmcli-charts#30.

## Why
External **networks** are auto-created on install; **secrets/configs** are validated (added in #418). Both were driven purely off the rendered manifest — a chart could not express *how* an external resource should be handled (auto-create vs operator-provisioned), nor give human-meaningful remediation. And on uninstall, auto-created networks were silently left behind (swarmcli could not tell "I created this" from "the user did").

## What

**#23 — requirements.yaml consumption**
- Load + validate `requirements.yaml` in both chart loaders (`chart.go`). Optional: `Chart.Requirements == nil` falls back to today's manifest-driven behaviour. New `Requirements`/`NetworkRequirement`/`ResourceRequirement` types; `*bool` `autoCreate`/`attachable` so an omitted key defaults `true` and an explicit `false` is distinguishable; `driver` defaults to `overlay`.
- The rendered manifest still **detects** the external set (values-aware — handles conditional networks). When `requirements.yaml` is present it is **authoritative**: `requireDeclared` rejects any external resource the manifest references that the file does not declare — a chart-authoring error, kept distinct from operator remediation.
- `autoCreate:true` networks are created with the declared driver/attachability; `autoCreate:false` networks are **validate-only** (a missing one is a hard error carrying the declared `description`, never auto-created). Descriptions also enrich secret/config remediation.
- Threaded via `InstallOptions.Requirements` (set by the CLI from `ch.Requirements`); no public `Install`/`Upgrade` signature change.

**#28 — symmetric uninstall summary** (decision on #28: option 1 — leave in place, write a summary)
- Persist the networks auto-created per revision on `Release.ManagedNetworks` (the previously-ephemeral `created[]`, made durable). Backward compatible — old records simply lack the field.
- `Uninstall` now returns `*UninstallResult{OrphanedNetworks}`: the union across revisions, filtered to those still present. Networks are **not** removed (they may be shared); the CLI reports them with `docker network rm` hints.

`Backend.CreateOverlayNetwork` gains `driver`/`attachable`.

## Tests
Parse/default, autoCreate/validate-only/contract-violation, `ManagedNetworks` persistence, uninstall reporting (+ skip-when-operator-removed), loader (dir/archive), and the no-requirements fallback — all via the `Backend` fake. Integration lifecycle test extended to exercise the requirements-driven auto-create + uninstall summary. `go test -race ./...` green; `go vet -tags=integration`, gofmt, and `golangci-lint run ./... --build-tags=integration` clean.

## Notes
No chart in swarmcli-charts requires changes beyond the docs/CI in the companion PR; the whoami chart already declares `traefik-public`.

### DIFF
--- a/charts/backend_docker.go
+++ b/charts/backend_docker.go
@@ -105,8 +105,8 @@ func (dockerBackend) NetworkScopes(ctx context.Context) (map[string]string, erro
 	return scopes, nil
 }
 
-func (dockerBackend) CreateOverlayNetwork(ctx context.Context, name string) error {
-	_, _, err := docker.CreateNetwork(ctx, name, network.CreateOptions{Driver: "overlay", Attachable: true})
+func (dockerBackend) CreateOverlayNetwork(ctx context.Context, name, driver string, attachable bool) error {
+	_, _, err := docker.CreateNetwork(ctx, name, network.CreateOptions{Driver: driver, Attachable: attachable})
 	return err
 }
 
--- a/charts/chart.go
+++ b/charts/chart.go
@@ -17,11 +17,12 @@ import (
 )
 
 const (
-	chartfileName = "Chart.yaml"
-	valuesName    = "values.yaml"
-	schemaName    = "values.schema.json"
-	readmeName    = "README.md"
-	templatesDir  = "templates"
+	chartfileName    = "Chart.yaml"
+	valuesName       = "values.yaml"
+	schemaName       = "values.schema.json"
+	readmeName       = "README.md"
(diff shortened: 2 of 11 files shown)


mosonyi:
Seems to be OK

PR #427, opened 2026-06-25
fix(charts): preserve comments in `charts show values`

eldara-cruncher:
## What

`charts show values <repo/chart>` re-marshalled the parsed values map, so it dropped the **comments** and original **key order** from the chart's `values.yaml`. Those comments are the chart's inline docs (defaults, required fields, gotchas), so losing them hurts.

## Fix

Retain the raw `values.yaml` bytes at load time (`Chart.ValuesRaw`, populated in both `LoadChartDir` and `LoadChartArchive`) and print them **verbatim** in `charts show values` — mirroring how `show schema` already emits raw `Schema` bytes. Falls back to marshalling the parsed map when a chart ships no `values.yaml`.

## Before / after (traefik chart)

```
# before: sorted keys, no comments
configs: []
deploy:
    placement:
        ...

# after: verbatim
# traefik chart default values.
#
# Deploys Traefik v3 as the Swarm edge proxy: ...
image:
  repository: traefik
  tag: ""  # defaults to appVersion from Chart.yaml
```

## Tests

- `charts`: `TestLoadChartDirRetainsValuesRaw`, `TestLoadChartDirNoValuesRawIsNil`, `TestLoadChartArchiveRetainsValuesRaw`
- `cli`: `TestChartsShowValuesPreservesComments` (verbatim output incl. comments)

`go test ./...`, `go vet`, `gofmt`, and `golangci-lint run` all clean.

Reported on Eldara-Tech/swarmcli-charts#37.

### DIFF
--- a/charts/chart.go
+++ b/charts/chart.go
@@ -53,6 +53,7 @@ func LoadChartDir(dir string) (*Chart, error) {
 		if err := yaml.Unmarshal(v, &ch.Values); err != nil {
 			return nil, fmt.Errorf("parse %s: %w", valuesName, err)
 		}
+		ch.ValuesRaw = v
 	} else if !os.IsNotExist(err) {
 		return nil, fmt.Errorf("read %s: %w", valuesName, err)
 	}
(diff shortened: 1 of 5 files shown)


mosonyi (approved):
LGTM

Issue #379, opened 2026-06-02
[STACK] healthcheck is missing


mosonyi:
At stack view (inspect) and edit, the healthcheck is missing

PR #429, opened 2026-06-25
fix(inspect): surface service logging driver in inspect/edit (#428)

### LINKED ISSUE #428: stack inspect/edit doesn't show logging
I have explicitly setup logging:

```
  logging:
    driver: loki:latest
    options:
      loki-url: http://loki:3100/loki/api/v1/push
```

But with inspect and edit it disappears

eldara-cruncher:
## What

Fixes #428 — a service deployed with a logging driver (e.g. Loki) lost its `logging:` block in both `stack inspect` and the `edit` (compose reconstruction) flows.

## Root cause

The Swarm `TaskTemplate.LogDriver` (`*swarm.Driver`: name + options) was dropped on the **render** side of both paths:
- `stack_to_compose.go` typed `TaskTemplate.LogDriver` as `any` (parsed but never read) and `ComposeService` had no logging field, so the `edit` reconstruction never emitted it.
- `stack_inspect.go` `ServiceSummary` had no logging field, so the inspect JSON never surfaced it.

## Fix

- New `Logging` (compose `driver:`/`options:`) and `LogDriver` (typed JSON capture) structs.
- `ReconstructStackCompose` and `GetStackInspection` now read `TaskTemplate.LogDriver` and emit `logging` via a shared `composeLogging()` helper — mirroring the existing `composeHealthcheck` pattern.
- Unset driver (`LogDriver: nil`, Swarm’s representation when none is pinned) emits **no** `logging:` block, so the daemon default is preserved — same "inherit ⇒ omit" behaviour as healthcheck.

## Tests

- Unit (`docker/stack_to_compose_test.go`): `composeLogging` variants, YAML emit/omit, and capture-side JSON unmarshal of `LogDriver` (both ends of the path).
- Integration: `logging:` assertions on `whoami_single` + a `json-file` driver fixture in `test-setup/test-stack.yml` (json-file so it deploys with no plugin), mirroring the #379 healthcheck integration pattern.

`go build` (both tags), `go test ./...`, `go vet`, `gofmt`, `golangci-lint` all clean.

### DIFF
--- a/docker/stack_inspect.go
+++ b/docker/stack_inspect.go
@@ -38,6 +38,7 @@ type ServiceSummary struct {
 	Configs     []string          `json:"configs,omitempty"`
 	Labels      map[string]string `json:"labels,omitempty"`
 	Healthcheck *Healthcheck      `json:"healthcheck,omitempty"`
+	Logging     *Logging          `json:"logging,omitempty"`
 	CreatedAt   time.Time         `json:"created_at"`
 	UpdatedAt   time.Time         `json:"updated_at"`
 }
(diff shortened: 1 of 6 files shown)


mosonyi:
LGTM

Issue #363, opened 2026-05-14
Edit Stack no network


mosonyi:
I added a new stack via swarmcli new stack in editor.

```yaml
version: "3.9"

services:
  nginx:
    image: nginx:latest

    ports:
      - "8080:80"

    deploy:
      replicas: 1
      restart_policy:
        condition: any

    networks:
      - default

networks:
  default:
    driver: overlay
```

When I wanted to edit, I got this screen:
```yaml
version: "3.8"
services:
    nginx:
        image: nginx:latest@sha256:06aa3d7be10bc6307990c81bdca075793132e9163391abc370c015e344e23128
        ports:
            - 8080:80/tcp
        deploy:
            replicas: 1
            restart_policy:
                condition: any
                max_attempts: 0
```

PR #431, opened 2026-06-25
fix(reconstruct): round-trip Tier-1/2 service-spec fields + completeness guard (#430)

### LINKED ISSUE #430: inspect/edit reconstruction silently drops many service-spec fields
## Summary

`#428` (logging dropped on `inspect`/`edit`) was one instance of a **systematic** gap: swarmcli's service→compose reconstruction is a hand-maintained, field-by-field allow-list with **no completeness guard**, so any service-spec field it does not explicitly map is silently dropped.

This matters most on the **`edit` path** (`docker/stack_to_compose.go` → `ReconstructStackCompose`), whose YAML is fed back to `docker stack deploy` — a dropped field is silent config loss on round-trip. The **`inspect` path** (`docker/stack_inspect.go` → `GetStackInspection`) is a human-readable summary, so most omissions there are by design (the one real gap is noted below).

## Two drop classes

- **Parse-drop** (worst): the field is absent from the intermediate structs (`ServiceInspect`/`TaskTemplate`/`ContainerSpec` in `stack_to_compose.go`), so it never even unmarshals — gone before any rendering runs.
- **Render-drop** (cheap to fix): the field is captured in the struct but the renderer never writes it.

### Tier 1 — common, dangerous to lose
- [ ] `deploy.update_config` (parallelism, delay, order, failure_action, monitor, max_failure_ratio) — **parse-drop** (no field on `ServiceSpec`)
- [ ] `deploy.rollback_config` — **parse-drop**
- [ ] `cap_add` / `cap_drop` (`ContainerSpec.CapabilityAdd/Drop`) — **parse-drop**
- [ ] `sysctls` — **parse-drop**
- [ ] `ulimits` — **parse-drop**
- [ ] `extra_hosts` (`ContainerSpec.Hosts`) — **parse-drop**
- [ ] `dns` / `dns_search` / dns options (`ContainerSpec.DNSConfig`) — **parse-drop**
- [ ] `hostname` (`ContainerSpec.Hostname`) — **parse-drop**
- [ ] `stop_grace_period` (`ContainerSpec.StopGracePeriod`) — **parse-drop**
- [ ] secret/config target `uid`/`gid`/`mode` — **render-drop** (parsed at `stack_to_compose.go:168-185`, only source/target written)
- [ ] `restart_policy.delay` / `restart_policy.window` — **render-drop** (parsed `:209/:211`, never written; only condition + max_attempts survive)

### Tier 2 — moderately common
- [ ] `read_only` (`ContainerSpec.ReadOnly`) — parse-drop
- [ ] `stop_signal` — parse-drop
- [ ] `init` — parse-drop
- [ ] `deploy.placement.preferences` (spread) — **render-drop** (parsed `:217`)
- [ ] `deploy.placement.max_replicas_per_node` — **render-drop** (parsed `:218`)
- [ ] `endpoint_mode` (vip/dnsrr, `EndpointSpec.Mode`) — parse-drop (struct has only `Ports`)
- [ ] generic/GPU resources (`Resources.GenericResources`) — parse-drop
- [ ] `pids_limit` (`Resources.*.Pids`) — parse-drop
- [ ] mount sub-options: tmpfs size/mode, bind propagation, volume nocopy/labels — render-drops (only `src:dst:ro` short-syntax emitted)

### Tier 3 — rare / verify applicability
- [ ] Groups, Privileges (cred_spec/selinux/seccomp/apparmor/no_new_privileges), Isolation, OomScoreAdj, TTY/stdin_open, replicated-job/global-job modes

### inspect-path correctness note
`GetStackInspection` emits only `svc.Spec.Labels` (service-level) at `stack_inspect.go:190`, never `ContainerSpec.Labels`, and does not run `filterLabels`, so it shows internal `com.docker.stack.*` keys (the edit path filters these). Minor, but worth aligning.

## Already covered (not drops)
image, env, command/args, ports, mounts/volumes, secrets/configs (names+target), cpu/mem limits & reservations, restart condition/max_attempts, placement constraints, replicas/global mode, networks, labels (both levels on edit path), healthcheck, and **logging** (`#428`).

## Recommended guard (prevent recurrence)
1. **Extract a pure core.** `ReconstructStackCompose` currently shells out to `docker service inspect` / `stack services` / `network ls` via `exec.Command` with no injection seam. Split out a pure `buildComposeService(si *ServiceInspect, stackName string, netID2Name map[string]string) ComposeService` (and/or whole-file `assembleCompose(...)`). This is the same seam any field fix needs anyway.
2. **Coverage test.** A "fat fixture → reconstruct → assert every set field present" table test in `docker/stack_to_compose_test.go`. A combined `TestReconstruct_DropsRichSpec` feeding one inspect JSON (restart delay/window, sysctls, cap_add, stop_grace_period, update_config, endpoint mode) is the failing baseline a fix flips to passing.
3. **Optional completeness check.** A reflection-based assertion that every non-trivial `ContainerSpec`/`TaskSpec` field is either rendered or explicitly allow-listed as intentionally dropped — turns future SDK additions into test failures instead of silent drops.
4. Note: there is currently **no `docker/stack_inspect_test.go`** — the inspect path is untested end-to-end.

## Repro without Docker
A unit fixture proves several drops at once: e.g. unmarshal an inspect JSON with `RestartPolicy.Delay` set and assert the rendered `deploy.restart_policy` has no `delay:` (render-drop); assert `ContainerSpec` has nowhere to hold `Sysctls`/`CapabilityAdd` (parse-drop); assert `EndpointSpec.Mode` is lost (parse-drop). Mirrors the existing `TestServiceInspect_CapturesLogDriver` (`stack_to_compose_test.go:440`).

Follow-up to #428. Audit done with code at current `main`.

eldara-cruncher:
## What

Fixes #430 — service→compose reconstruction was a hand-maintained allow-list with **no completeness guard**, so any unmapped service-spec field was silently dropped. On the **`edit` path** the reconstructed YAML is fed back to `docker stack deploy`, so a dropped field is **silent config loss on round-trip** (including security-relevant ones: caps, secret file mode, `read_only`).

Scope (agreed): **Tier-1 + cheap Tier-2** fields, the inspect label-filter alignment, and a recurrence guard. Involved Tier-2 and Tier-3 are deferred.

## Changes

**Seam (no behaviour change).** Extract a pure `assembleCompose()` + `(*ComposeFile).buildService()` out of `ReconstructStackCompose`; the four `declare*` closures become methods. All `exec` stays in `ReconstructStackCompose`, so the assembler is unit-testable without Docker.

**Tier-1 fields mapped:** `update_config` / `rollback_config`, `cap_add` / `cap_drop`, `sysctls`, `ulimits`, `extra_hosts` (Swarm `"IP host"` → compose `"host:IP"` reorder), `dns` / `dns_search` / `dns_opt`, `hostname`, `stop_grace_period`, secret/config target `uid`/`gid`/`mode`, `restart_policy.delay` / `.window`.

**Cheap Tier-2 mapped:** `read_only`, `stop_signal`, `init`, `pids_limit` (→ `deploy.resources.limits.pids`), `endpoint_mode`, `placement.max_replicas_per_node`.

All new fields are gated on non-zero/non-nil so reconstruction never pins a Swarm default the user never wrote (cf. #363). Durations render via `time.Duration(ns).String()` (same as healthcheck). Secret/config `mode` is emitted as the decimal permission value (`0o444` == `292`) for unambiguous round-trip through the compose loader.

**Inspect path (bugfix, C0):** run `filterLabels` on service labels and surface a new `container_labels` field, so internal `com.docker.stack.*` keys no longer leak into inspect JSON (aligns with the edit path).

## Guard against recurrence

- **Always-on:** `TestReconstruct_DropsRichSpec` — one fat `docker service inspect`-shaped JSON exercising every supported field → `assembleCompose` → assert each survives (catches both parse- and render-drops). Plus per-field unit tests and the first-ever `docker/stack_inspect_test.go`.
- **Integration:** `whoami_single` / `default_net_probe` fixtures in `test-stack.yml` carry the new fields; `TestReconstructStackCompose`, the `_RoundTrip` test, and the inspect test assert they survive a real deploy → inspect → reconstruct.
- **Opt-in advisory** (`//go:build sdkcoverage`, **not** in the default CI matrix): reflects over `swarm.ContainerSpec`/`TaskSpec`/`ServiceSpec` and fails if a field is neither mapped nor in an explicit `intentionallyDropped` allow-list — run it when bumping the Docker SDK to surface newly-added fields.

## Deferred (tracked in the advisory allow-list)

- Involved Tier-2: mount long-syntax sub-options, generic/GPU resources, `placement.preferences`.
- Tier-3: Groups, Privileges, Isolation, OomScoreAdj, TTY/stdin, job modes.

## Tests

`go build` (both tags), `go test ./...`, `go test -tags sdkcoverage ./docker/...`, `go vet`, `gofmt`, `golangci-lint --build-tags=integration` all clean. Integration tests run against the DinD Swarm via CI.

## Coordination

Sibling of the still-open **#429** (logging, #428): both touch `docker/stack_to_compose.go` and `test-stack.yml`. This PR branches off `main` and is independent; whichever merges second resolves the textual overlap (the logic is disjoint — `LogDriver` is listed as handled-by-#429 in the advisory).

### DIFF
--- a/docker/stack_coverage_sdk_test.go
+++ b/docker/stack_coverage_sdk_test.go
@@ -1,0 +1,114 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+//go:build sdkcoverage
+
+// Opt-in advisory: build/run with `-tags sdkcoverage`. It is intentionally
+// kept OUT of the default CI matrix so a routine Docker SDK bump does not
+// red-bar unrelated PRs. Run it when bumping github.com/docker/docker to
+// surface swarm spec fields that were added upstream but are not yet mapped
+// (or explicitly waived) by the service→compose reconstruction (see #430).
+//
+//	go test -tags sdkcoverage ./docker/...
+
+package docker
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/docker/docker/api/types/swarm"
+)
+
+// TestSDKFieldCoverage fails when a field on one of the reconstructed swarm
+// specs is neither mapped by the reconstruction nor explicitly listed as
+// intentionally dropped. This turns a future silent parse-drop into a loud,
+// reviewable test failure: the contributor bumping the SDK must decide to map
+// the new field or waive it here.
+func TestSDKFieldCoverage(t *testing.T) {
+	cases := []struct {
+		name string
+		typ  reflect.Type
+		// covered maps an SDK field name to either "handled" (mapped into the
+		// compose model) or "dropped: <reason>" (a conscious omission).
+		covered map[string]string
+	}{
+		{
+			name: "ContainerSpec",
+			typ:  reflect.TypeOf(swarm.ContainerSpec{}),
+			covered: map[string]string{
+				"Image":           "handled",
+				"Labels":          "handled",
+				"Command":         "handled",
+				"Args":            "handled",
+				"Hostname":        "handled",
+				"Env":             "handled",
+				"Dir":             "handled",
+				"User":            "handled",
+				"Init":            "handled",
+				"StopSignal":      "handled",
+				"ReadOnly":        "handled",
+				"Mounts":          "handled",
+				"StopGracePeriod": "handled",
+				"Healthcheck":     "handled",
+				"Hosts":           "handled",
+				"DNSConfig":       "handled",
+				"Secrets":         "handled",
+				"Configs":         "handled",
+				"Sysctls":         "handled",
+				"CapabilityAdd":   "handled",
+				"CapabilityDrop":  "handled",
+				"Ulimits":         "handled",
+				"Groups":          "dropped: Tier-3, no common compose mapping",
+				"Privileges":      "dropped: Tier-3 (cred_spec/selinux/seccomp/apparmor/no_new_privileges)",
+				"TTY":             "dropped: Tier-3 (tty)",
+				"OpenStdin":       "dropped: Tier-3 (stdin_open)",
+				"Isolation":       "dropped: Tier-3 (windows isolation)",
+				"OomScoreAdj":     "dropped: Tier-3",
+			},
+		},
+		{
+			name: "TaskSpec",
+			typ:  reflect.TypeOf(swarm.TaskSpec{}),
+			covered: map[string]string{
+				"ContainerSpec":         "handled",
+				"Resources":             "handled",
+				"RestartPolicy":         "handled",
+				"Placement":             "handled",
+				"Networks":              "handled",
+				"LogDriver":             "handled",
+				"PluginSpec":            "dropped: non-container runtime",
+				"NetworkAttachmentSpec": "dropped: non-container runtime",
+				"ForceUpdate":           "dropped: internal counter, not user config",
+				"Runtime":               "dropped: non-container runtime",
+			},
+		},
+		{
+			name: "ServiceSpec",
+			typ:  reflect.TypeOf(swarm.ServiceSpec{}),
+			covered: map[string]string{
+				"Annotations":    "handled", // embedded: Name + Labels
+				"TaskTemplate":   "handled",
+				"Mode":           "handled",
+				"UpdateConfig":   "handled",
+				"RollbackConfig": "handled",
+				"Networks":       "handled",
+				"EndpointSpec":   "handled",
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		for i := 0; i < tc.typ.NumField(); i++ {
+			f := tc.typ.Field(i)
+			if f.PkgPath != "" {
+				continue // unexported
+			}
+			if _, ok := tc.covered[f.Name]; !ok {
+				t.Errorf("swarm.%s.%s is not mapped or waived — map it in "+
+					"stack_to_compose.go or add it to the dropped set (#430)",
+					tc.name, f.Name)
+			}
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


eldara-cruncher:
Rebased onto `main` now that #429 (logging) is merged. Logging is folded into the new seam rather than left as a sibling:
- `LogDriver` typed + `Logging` struct + `composeLogging` are re-expressed inside `buildService` / the inspect path (alongside `container_labels`).
- `LogDriver` flipped from `dropped` to `handled` in the sdkcoverage advisory.
- The fat-fixture (`TestReconstruct_DropsRichSpec`) and `whoami_single` integration fixture now also carry `logging:`, so the general guard covers it too.
- Also includes the cap-name fix (Docker canonicalizes `SYS_NICE`→`CAP_SYS_NICE`).

Local: build (both tags), `go test ./...`, `-tags sdkcoverage`, vet, gofmt, golangci-lint all clean.

mosonyi:
LGTM: The logging still works.

eldara-cruncher:
Moved #430's field coverage to **unit-only** and reverted the `test-stack.yml` + integration assertions to `main`.

Investigation conclusion: the integration *fixture* additions to `whoami_single`/`default_net_probe` deterministically perturbed the shared demo stack enough to stall the 3rd back-to-back `demo_whoami` restart in `TestRestartServiceWithProgress/multi_replica` (confirmed: #432 — de-apk, no fixture — passed; this branch with the fixture failed across runs). The reconstruction code and the round-trip test are not implicated (round-trip finishes ~70s before the failure).

Every field remains rigorously covered by the unit fat-fixture (`TestReconstruct_DropsRichSpec`: capture JSON → reconstruct → assert), the per-field unit tests, and the `sdkcoverage` advisory. Integration now matches `main`, so the restart tests pass. Net result: the production fix + its guard are intact; only the cluster-perturbing confirmatory fixture is gone.

PR #437, opened 2026-07-02
feat(services): container health + port in the services view

eldara-cruncher:
## What

Adds the plumbing + rendering for **container health** and **per-replica port** in the services view. The remote Docker/Swarm API cannot report container health (`Up 8m (healthy)`), so today the view can't show whether replicas are actually healthy.

- `docker.TaskEntry.Health` and `docker.ServiceEntry.Health` fields — **generic extension points**. The default (remote-API) loaders leave them empty; they are populated by a `docker.Deps` decorator that can reach per-node container state.
- Expanded per-replica rows gain **conditional HEALTH + PORTS columns** and tint unhealthy replicas red / starting yellow (mirrors the existing service-level error coloring).
- The services list gains a **conditional HEALTH summary column** (after STATUS), shown only when a row carries a summary.

Everything is invisible in plain CE (fields stay empty) — no behavior change without a decorator supplying data.

## Why

Foundation for Business Edition issue Eldara-Tech/swarmcli-be#236 ("Better view on services"). The BE side sources health from the on-node agent and injects it through these fields.

## Tests

Added `columns_test.go` coverage: HEALTH column appears only when populated (and sits after STATUS); `formatTaskRow` omits HEALTH/PORTS unless present. Full suite + `golangci-lint` green.

## Also: discoverability footer hint (services view)

Adds a one-line services-view footer hint so the HEALTH/PORTS capability is discoverable, mirroring the volumes footer-flag pattern (`views/volumes/view.go:78-87`): shown while the `service-health` flag is off (CE / unlicensed → upsell to `view.BELandingURL`), replaced by the HEALTH column when a BE build enables the flag. Also adds a generic `view.ServicesHealthHint` seam so the BE module can surface a context-specific note without CE knowing about managed contexts. Pro-boundary-safe (CE names only a capability + the landing URL).

Follow-up: rather than dropping the HEALTH column when there's no live data, it now stays visible with a `*` placeholder whenever a footnote applies (CE upsell / non-managed context), and the footer note is prefixed with `* ` — so the missing capability is discoverable in-column, not only in the footer. The column is dropped only when health is reachable and no service reports any.

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -356,8 +356,14 @@ type ServiceEntry struct {
 	Mode           string
 	Image          string
 	Ports          string
-	CreatedAt      time.Time
-	UpdatedAt      time.Time
+	// Health is an aggregate health summary for the service's running replicas
+	// (e.g. "2/2 healthy"); "" when unknown. The swarm API does not expose
+	// container health, so the default loaders leave it empty; it is an
+	// extension point populated by a ServiceOps decorator that can reach
+	// per-node container state.
+	Health    string
+	CreatedAt time.Time
+	UpdatedAt time.Time
 }
 
 func LoadNodeServices(nodeID string) []ServiceEntry {
--- a/docker/task.go
+++ b/docker/task.go
@@ -29,9 +29,17 @@ type TaskEntry struct {
 	// the status is unknown. The swarm task snapshot does not carry it, so the
 	// default loaders leave it empty; it is an extension point populated by a
 	// TaskOps decorator that can reach per-node container state.
-	Health    string
-	CreatedAt time.Time
-	UpdatedAt time.Time
+	Health string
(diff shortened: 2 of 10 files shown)


eldara-cruncher:
**Code review follow-up (e-review)** — commit 013138a.

(1) Truncate the expanded-row HEALTH cell via `TruncateRunes(…, 9)` — it was the only sub-row cell not length-capped, so an over-long value could jag PORTS/ERROR against the fixed-width header. (2) Guard the `features` map with an `RWMutex`: it was read from tea.Cmd goroutines and written from the update goroutine (`ApplyLicenseState`) — a latent concurrent-map panic the new service-health readers amplify. Review otherwise confirmed plain-CE output is unchanged (fields default empty ⇒ columns hidden) and the pro boundary is respected.

PR #442, opened 2026-07-12
feat(tasks): PullProgress extension point on task and service rows

eldara-cruncher:
## Why

A task whose image is still downloading sits in `preparing` for as long as the pull takes, and the TUI can say nothing more than that. The Swarm API carries **no** image-pull progress at all — the swarm executor consumes the pull's progress stream into a debug log line (moby `daemon/cluster/executor/container/adapter.go`) — so nothing the manager can query knows how far along a multi-gigabyte pull is. Deploying something like `ollama` therefore looks indistinguishable from a hang.

The bytes are only observable on the node doing the pull. This PR adds the CE half: the display seam. The observation itself lives in the agent, and the decorator that fills these fields lives in the BE module.

## What

Two extension-point fields, in the same shape as the existing `Health` / `ContainerState` fields:

- `docker.TaskEntry.PullProgress` — the pull in flight for that task.
- `docker.ServiceEntry.PullProgress` — the same, summarized for the service row.

The default loaders leave both empty, so **base CE behaves exactly as before**. A `TaskOps` / `ServiceOps` decorator that can reach the node performing the pull fills them in.

`TaskEntry.StatusText()` is the accessor the status cell now goes through, so a pulling task shows the pull instead of a bare `preparing`; the services `STATUS` column likewise shows it instead of a bare `active`.

```
SERVICE   REPLICAS  STATUS         IMAGE
ollama    0/1       pulling 41%    ollama/ollama:latest
  \_ ollama.1                      pulling 41% · 1.2 GB / 2.9 GB
```

No new columns, no layout change, no feature flag in CE (the fields are simply empty when nothing fills them).

## Notes for review

- **The tasks view's cell is switched too, but is inert today.** `views/tasks` discards `Deps` (`factory(_ docker.Deps, …)`) and calls the package-level loaders, so no decorator runs there. The same pre-existing gap already leaves `Health` empty in that view. Using `StatusText()` there keeps the three status cells consistent and makes the view correct for free if `Deps` is ever threaded through; it is not a fix for that gap.
- **Sorting by STATUS still sorts on `ServiceEntry.Status`**, not the rendered cell, so a transient pull does not reshuffle the list.
- Pro Feature Boundary respected: the fields are documented as generic extension points, with no pro naming or internals.

Refs Eldara-Tech/swarmcli-be#250

### DIFF
--- a/docker/service.go
+++ b/docker/service.go
@@ -361,9 +361,16 @@ type ServiceEntry struct {
 	// container health, so the default loaders leave it empty; it is an
 	// extension point populated by a ServiceOps decorator that can reach
 	// per-node container state.
-	Health    string
-	CreatedAt time.Time
-	UpdatedAt time.Time
+	Health string
+	// PullProgress summarizes an image pull in flight for this service's tasks
+	// (e.g. "pulling · 3/12 layers · 412 MB"); "" when nothing is being pulled or
+	// the progress is unavailable. Like Health it is an extension point populated
+	// by a ServiceOps decorator that can reach the nodes performing the pull; the
+	// services view shows it in place of Status while it is set, since a service
+	// whose image is still downloading otherwise reads as a bare "active".
+	PullProgress string
+	CreatedAt    time.Time
+	UpdatedAt    time.Time
 }
 
 func LoadNodeServices(nodeID string) []ServiceEntry {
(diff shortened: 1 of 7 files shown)


mosonyi:
Why do we need this in the CE? I tested BE , all good. If you think it is needed, you can close.

clangenb:
It does work on local nodes with this extension, and it also adds it to the general type definitions of our docker data types that are also used by BE

PR #454, opened 2026-07-14
feat(views): add Chromeless opt-in seam for full-terminal views

eldara-cruncher:
## What

Adds an opt-in `Chromeless` view interface. When a view reports `Chromeless() == true`, the app renders its `View()` verbatim — no help bar, no view frame, no breadcrumb bar — and hands it the **full terminal size** instead of the chrome-reduced viewport. App-level overlays (startup, unlock, error, update) still composite on top.

## Why

Views return frame components and `app/view.go` wraps them in the chrome. A view that needs the raw terminal — an embedded terminal emulator, say, where the frame's `│` lands in the user's clipboard on copy — has no way to opt out. This is the generic extension point for that; no view in this repo uses it yet.

`m.currentView.View()` was dead code (the app only ever called the four `Frame*` methods), so this repurposes an existing interface method rather than adding one.

## Also in here

Three key-handling guards, each a correctness fix rather than polish:

- **`:`** would open a command bar that a chromeless view never renders — and an *invisible* command bar then swallows every subsequent keypress, since all keys are routed to it exclusively. Chromeless views now go straight to the key handler.
- **`esc`** while `m.fullscreen` happens to be set would invisibly clear the flag instead of going back, costing the user a second `esc` to leave the view.
- **`f`** has nothing to hide over an already-borderless view, so the key belongs to the view.

The four resize call sites are consolidated into `resizeView`/`resizeCurrentView`. That also fixes a pre-existing bug: `switchToView`/`replaceView`/`goBack` all resized the new view with a hardcoded `isFullscreen=false`, so entering a view while in fullscreen left the restored view six lines short on the way back.

`m.viewport` deliberately stays chrome-sized — `goBack` and the command/search-bar paths derive the *restored* view's dimensions from it.

## Contracts (documented on the interface)

- `Chromeless()` must be **constant** for the view's lifetime: it selects both the chrome and the size handed to the view, and there is no re-resize hook on the render path.
- `View()` must emit **exactly `terminalHeight` lines**, no trailing newline. Bubble Tea drops surplus lines from the *top*, so a one-line overshoot silently eats the view's first row.

## Test plan

- `views/view/interface_test.go` — `IsChromeless` for opting-in, opting-out and non-implementing views.
- `app/chromeless_test.go` — a chromeless view renders verbatim (no `│`); gets the raw terminal size while `m.viewport` stays chrome-sized; `:` / `/` do not open their bars; `f` does not toggle fullscreen; `esc` goes back even when `m.fullscreen` is set.
- `go test ./...` and `golangci-lint run ./... --build-tags=integration` clean.

First consumer is swarmcli-be's exec shell (Eldara-Tech/swarmcli-be#258).

### DIFF
--- a/app/chromeless_test.go
+++ b/app/chromeless_test.go
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package app
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/stretchr/testify/require"
+)
+
+// A chromeless view is rendered verbatim: no help bar, no frame, no breadcrumbs.
+func TestChromelessView_RenderedVerbatim(t *testing.T) {
+	m := newTestAppModel(&chromelessStubView{})
+
+	out := m.View()
+
+	require.Equal(t, "CHROMELESS", out)
+	require.NotContains(t, out, "│")
+}
+
+// A chromeless view gets the raw terminal size, while m.viewport stays
+// chrome-sized so views further down the stack are restored correctly.
+func TestChromelessView_GetsFullTerminalSize(t *testing.T) {
+	cv := &chromelessStubView{}
+	m := newTestAppModel(cv)
+
+	m.updateForResize(tea.WindowSizeMsg{Width: 100, Height: 40})
+
+	require.Equal(t, []tea.WindowSizeMsg{{Width: 100, Height: 40}}, cv.receivedSizes)
+	require.Equal(t, 96, m.viewport.Width)
+	require.Equal(t, 40, m.viewport.Height)
+}
+
+// A framed view still gets the chrome-reduced size (help bar + breadcrumb bar).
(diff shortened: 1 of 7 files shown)


eldara-cruncher:
Closing: no longer needed.

The consumer this seam was built for — the exec shell in swarmcli-be#258 — no longer renders inside the TUI at all. Review feedback asked for native terminal scrollback (mouse-drag past the top edge should scroll), and a Bubble Tea view cannot provide that at any level of borderlessness: the program lives in the alt screen, which has no scrollback.

The shell now hands the real terminal over via `tea.Exec` instead — the same path this repo already uses for `$EDITOR` — so `Chromeless` has no consumer left, and merging an unused extension point into CE would just be dead surface area.

One thing in here was a genuine, independent bug and is worth keeping in mind: `switchToView`/`replaceView`/`goBack` all resize the new view with a hardcoded `isFullscreen=false`, so navigating while in `f` fullscreen leaves the restored view six lines short. That is still present on main; happy to send it as its own small PR if you want it.

PR #455, opened 2026-07-15
UI: recolor & reformat service/stack view headers (#453)

### LINKED ISSUE #453: Change Header in views
##Service View:
In service view there's a header like:
---- Services in Stack: postgres-ha ----
Change it to
---- Services(<stakname>)[<count of services>] -----
for instance:
---- Services(postgres-ha)[8] -----

Code colors:
Services: #8be4e4 
<stackname>: #ff04ff
<number>: #ffefd5

## Stack view
Stack on Node (Total: 38) 
change to:
stacks (all)[38]

stacks: #8be4e4 
all: #ff04ff
38:  #ffefd5

eldara-cruncher:
## What
Reformats the service-view and stack-view frame titles into a coloured `Label(scope)[count]` form (issue #453):

- Service view: `Services in Stack: postgres-ha` → `Services(postgres-ha)[8]`
- Stack view: `Stacks on Node (Total: 38)` → `Stacks(all)[38]`

Per-token colours: label + separators cyan `#8be4e4`, scope magenta `#ff04ff`, count peach `#ffefd5`.

Per discussion on the issue, **all four** service-view title variants are harmonized to the same format for consistency:

| Filter | Header |
|---|---|
| stack | `Services(<stack>)[n]` |
| node | `Services(<nodeID>)[n]` |
| no-stack | `Services(no stack)[n]` |
| all | `Services(all)[n]` |

## How
- New `ui.ScopedTitle(label, scope, count)` builds the per-segment-coloured title.
- `ui.styleFrameTitle` teaches the frame renderer to pass an already-styled (ANSI-containing) title through unchanged instead of re-wrapping it with `FrameTitleStyle`. Without this, lipgloss's inner colour resets clobber the outer style on the separators. Plain titles from every other view are unaffected (backward-compatible).

## Tests
- New `ui/title_test.go`: title format across variants, per-token colour wiring, and the pass-through guard.
- Updated the two view tests that asserted the old header strings.
- Full suite + `golangci-lint --build-tags=integration` pass; verified end-to-end under a forced truecolor profile that the segment colours survive into the rendered border.

Closes #453

### DIFF
--- a/ui/framedbox.go
+++ b/ui/framedbox.go
@@ -15,7 +15,7 @@ import (
 // line and raw content are returned (no borders).
 func RenderViewFrame(title, header, content, footer string, width, height int, fullscreen bool) string {
 	if fullscreen {
-		titleText := FrameTitleStyle.Render(title)
+		titleText := styleFrameTitle(title)
 		titleLine := lipgloss.NewStyle().
 			Width(width).
 			Align(lipgloss.Center).
--- a/ui/title.go
+++ b/ui/title.go
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package ui
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/lipgloss"
+)
+
+// Frame-title token styles. A scoped title reads "Label(scope)[count]"; the
+// label and its surrounding separators share one colour so the title reads as
+// a single unit, while the scope and count are highlighted distinctly.
+var (
+	titleLabelStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#8be4e4")).Bold(true)
+	titleScopeStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("#ff04ff")).Bold(true)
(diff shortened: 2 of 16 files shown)


mosonyi:
If I filter on the stacks/services it should be ajusted as well

stacks(filtername)[1]

image

eldara-cruncher:
Good catch — done in fa63585.

Implemented the k9s way: when a `/` filter is active the header **count now reflects the filtered rows**, and the filter is **appended** as a `</query>` fragment rather than folded into the scope. Your example now renders:

```
Stacks(all)[1] </pos>
```

k9s keeps the scope in the parens and appends the search term (e.g. `Pods(all)[1] </pos>`), so I kept `(all)` and added `</pos>` instead of replacing the scope with the filter text as in `stacks(filtername)[1]`. Happy to switch to the literal `Stacks(pos)[1]` form if you'd rather.

Applied to **stacks, services and nodes** (nodes had the same total-vs-filtered count bug); volumes/networks/secrets/configs already counted filtered rows. The filter text is truncated in the header so a long query can't overflow the title.

mosonyi:
perfect, it can be closed.

PR #460, opened 2026-07-17
Show ":help" hint on the stack bar, clamp breadcrumbs to width

### LINKED ISSUE #459: type :help
Add a small info here:

Type :help for  commands

Image

eldara-cruncher:
Closes #459.

## What

Adds a muted `Type :help for commands` to the **breadcrumb bar** (the last line of the screen), right-aligned.

The `:` command bar is invisible until pressed (`commandinput.View()` returns `""` when inactive), and neither help surface advertises it — the helpbar's `<?> Help` opens the *keybinding cheat-sheet*, while the *command list* lives behind `:help`. So `:stack`, `:service`, `:node` … were effectively undiscoverable.

```
 stacks                                    Type :help for commands
```

## Two deviations from the issue worth reviewing

**1. Placement is one row below the screenshot.** The red box in #459 sits on the `Stack 1 of 10` row *inside* the frame border — the view's `FrameFooter`. Putting the hint there costs `logs`/`inspect`/`contexts`/`loading` one content row each: they return `""` from `FrameFooter()`, so a hint flips `footerLines` 0→1 in `ui/framecalc.go:57` (14→13 rows on the log viewer at 80×24, where vertical space matters most). It would also mean `app/view.go` doing string surgery on view-owned, already-styled footers plus duplicating `borderWidth` math that `ui/` owns.

The breadcrumb bar is one row lower, costs **zero** rows (the line already exists and is already budgeted by `view.go:56`'s `-1` and `update.go:468`'s `-7`), and needs no layout arithmetic changes at all.

**2. A pre-existing bugfix is folded in** (at the reviewer's request). Breadcrumbs were never width-clamped, so a nested trail rendered *wider than the terminal* below ~40 columns — `model.go:193` returned them unclamped. `fitBreadcrumbs` sheds the oldest segments until the line fits, reusing the `…` collapse `RenderBreadcrumbs` already implements, so **the current view always survives** (a plain right-truncation would eat exactly the crumb that matters most). `MaxWidth` backstops the pathological case.

`TestRenderStackBar_NeverExceedsTerminalWidth` **fails on `main`** — verified: the trail renders 34 columns into a 10-column terminal. That failure is the bugfix landing. It also turns `renderStackBar`'s `gap` arithmetic from sound-by-assumption into sound-by-construction.

## Extension-point safety

`StackBarSuffix` is **read, never written**. The hint composes as a separate segment to its left; the suffix keeps its documented right-edge anchor (no jitter when the hint drops), and the hint is dropped first when the line is narrow — persistent status outranks a discoverability aid. Rendering is a strict superset of today's: byte-identical when the hint doesn't fit, gaining the hint only when there is room.

Verified against the BE module (which has `replace swarmcli => ../swarmcli`): full suite green, including the four packages that assert on `StackBarSuffix`.

```
 stacks       Type :help for commands  No valid license · Community Edition   (w=80)
 stacks                                No valid license · Community Edition   (w=60, hint sheds)
```

## Suppression

Hidden in the help view (`:help` is what opens it — `?` lands on the same view name, so one check covers both) and under `HidesGlobalHelp()`, where the view captures every keystroke and `:` genuinely does not work — advertising it there would be a lie. Reuses the check the helpbar already needed; `hidesGlobalKeys()` is extracted rather than duplicated (net −4 lines in `view.go`).

Fullscreen needs no handling: `view.go:16-23` returns before `renderStackBar` is called.

## Wording

"Type", not "Press" — the prompt glyph renders `> `, so there is no `:` on screen to press. "for commands", not "for help" — `?` shows keybindings, `:help` shows the command list; the wording is what keeps them distinct.

## Test / verification

`app/stackbar_test.go` is net-new coverage — `renderStackBar` had **zero** tests. 9 tests: hint shown, BE coexistence + ordering, degradation order, both suppressions, the three clamp cases, and a width sweep (10→200 × 3 suffix lengths) asserting the bar never exceeds `terminalWidth` and stays one line.

`gofmt` / `go vet` / `golangci-lint` (0 issues) / `go test -race -count=1 ./...` all green in both repos. Rendered output eyeballed at widths 10–120 across OSS, BE-unlicensed, and both suppressed surfaces.

## Labels

`A0-ui` even though the clamp is a bugfix — group A takes one label and the headline intent is the hint.

### DIFF
--- a/app/model.go
+++ b/app/model.go
@@ -174,25 +174,85 @@ func (m *Model) replaceView(name string, data any) tea.Cmd {
 // Set from init() to display persistent status (e.g., license mode).
 var StackBarSuffix string
 
+// commandHintText advertises the ":" command bar, which is invisible until
+// pressed. ":help" opens the command list; "?" opens the keybindings — the
+// wording deliberately keeps those two distinct.
+const commandHintText = "Type :help for commands"
+
+// stackBarGap is the minimum number of blank columns kept between the
+// breadcrumbs and the right-aligned block, and between the segments inside it.
+const stackBarGap = 2
+
+// stackBarMaxCrumbs is the most breadcrumb segments shown before the oldest
+// are collapsed into a "…" prefix.
+const stackBarMaxCrumbs = 3
+
+var (
+	stackBarSuffixStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("208"))
+	stackBarHintStyle   = lipgloss.NewStyle().Foreground(lipgloss.Color("240"))
+)
+
 func (m *Model) renderStackBar() string {
 	names := make([]string, 0, m.viewStack.Len()+1)
 	for _, v := range m.viewStack.Views() {
 		names = append(names, v.Name())
 	}
 	names = append(names, m.currentView.Name())
-	result := RenderBreadcrumbs(names, 3)
-	if StackBarSuffix == "" {
-		return result
+	crumbs := m.fitBreadcrumbs(names)
+
+	// Try the right-aligned renderings widest-first and take the first that
+	// fits. The hint is dropped before the suffix: the suffix carries
+	// persistent status, the hint is only a discoverability aid.
+	for _, right := range m.stackBarRight() {
+		gap := m.terminalWidth - lipgloss.Width(crumbs) - lipgloss.Width(right)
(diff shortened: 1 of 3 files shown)


mosonyi (approved):
LGTM

PR #461, opened 2026-07-17
feat(charts): let a chart declare the swarmcli it needs

eldara-cruncher:
## Why

Charts increasingly depend on chart-engine behaviour newer than the swarmcli a user has installed — `swarmcli-charts/scripts/install-swarmcli.sh` builds CE from `main` precisely because it tracks unreleased features. There was no field in which a chart could say so, so installing one on an older build failed at render time with whatever the missing feature happened to produce:

```
template: stack.yaml:14: function "toYamlPretty" not defined
```

which leaves "my swarmcli is too old" indistinguishable from "this chart is broken".

## What

A chart may now declare a SemVer constraint on the chart engine, modelled on npm's `engines` (Helm has no `helmVersion` field — this is a gap in Helm, not a pattern copied from it):

```yaml
# Chart.yaml
swarmcliVersion: ">= 1.13.0"
```

```
$ swarmcli charts install web ./traefik
chart traefik 0.2.0 requires swarmcli >= 1.13.0; this build provides 1.12.0

Continue anyway? [y/N]
```

Detection is a pure function in `charts/` (`CheckCompat`); policy lives in `cli/`, mirroring the existing `RepoStore.Warnf` split — so `charts` stays testable without a terminal.

| Verb | Behaviour |
|---|---|
| `install`, `upgrade` | refuse; prompt `[y/N]` (default **No**) when interactive, abort when not |
| `apply` | refuse, **never** prompts — it is meant to run unattended and must not block on stdin |
| `apply --dry-run` / `--diff`, `template`, `diff`, `show` | warn only |
| any | `--skip-compat-check` forces through |

Nothing declared, an unstamped build, or an unparseable constraint all yield `CompatUnknown` and **never block** — mirroring the doctrine already in `swarmcli-be/internal/compat`: an unknown version "is not evidence of an incompatible stack". This is a compatibility aid, not a security boundary (a chart already renders to an arbitrary stack), so it fails open on our own inability to parse rather than breaking working charts for a cosmetic reason.

### The engine version, not `main.version`

The constraint is checked against the **chart engine's** version, stamped via `-X swarmcli/charts.engineVersion`, *not* `main.version`. The engine is this module's code, so a binary embedding it carries whichever engine it pinned regardless of its own tag — comparing against the binary's own version would be silently wrong there. When the two differ the diagnostic names both, so the number cited is one the reader can map to something they installed:

```
chart traefik 0.2.0 requires swarmcli >= 1.13.0; this build provides chart engine 1.12.0 (swarmcli 1.12.1)
```

Release candidates are compared on their **core** version: SemVer constraints exclude prereleases, so `>= 1.13.0` would otherwise reject `1.13.0-rc1` — firing exactly while a release is under test.

## Chart.yaml `apiVersion` is now validated

`Chartfile.APIVersion` was parsed but never validated — `apiVersion: v99` loaded fine. It is now allowlisted to `{"", "v1"}`, matching `releasefile.go:87` and `generate-index.sh`, which reserves `v2`+ for a real format break. This is the structural gate Helm's own `apiVersion` provides, and the reason a future format change can make older builds refuse a chart they cannot read.

This narrows the Chart.yaml contract — a chart declaring `apiVersion: v2` is now refused — so it deserves an explicit note on why it is labelled **C0-breaks-nothing** rather than C1:

- **No user is affected.** None of the 12 published charts set `apiVersion` (0/12), and `swarmcli-charts` is the ecosystem today. `new-chart.sh` does not scaffold it either.
- **The only `v2` in existence was ours.** `charts/testdata/demo/Chart.yaml` and both `integration-tests` fixtures said `v2` — Helm-3 muscle memory that arrived unexamined in #415 and survived only because nothing validated it, while the *same* PR's release-file parser enforces `v1`. Those three migrate to `v1` here.
- **Weighed against the alternative.** C1 would force the next GA to `v2.0.0` (current: v1.12.0), signalling a migration to users when there is none, and would hold `swarmcliVersion` — whose whole value is forward-looking — behind a major release.

Flagging it plainly since the label is this repo's only breaking-change gate (the changelog is type-only): if you would rather record the contract change formally, relabel to C1 and this ships as `v2.0.0` with `.github/UPGRADE_NOTES.md` filled.

## Known limitation (by design, stated not engineered around)

`Chart.yaml` is parsed leniently, so **every already-released swarmcli silently ignores `swarmcliVersion`**. Only builds carrying this check can honour it — the same bootstrapping limit Helm's `apiVersion` gate has. Making `Chart.yaml` strict would not fix it and *would* break the published charts, which ship `annotations`/`keywords`/`home`/`sources` that `Chartfile` does not model. The value is forward-looking, so the sooner this lands the sooner the floor moves.

## Verification

Beyond `go test ./...`, `-race` on `charts`/`cli`, and `golangci-lint --build-tags=integration` (all clean), the built binary was driven end-to-end:

- read-only verbs warn and exit 0; `install` refuses **before touching Docker**; `--skip-compat-check` forces through; an unknown flag still fails loudly
- interactive prompt on a real pty: `n` and a bare Enter both abort (exit 1), `y` proceeds through to Docker
- unstamped `go build` warns and proceeds; a satisfied constraint is completely silent
- `apiVersion: v99` refused; **all 12 published charts still load**
- binary/engine divergence stamped separately renders the two-version message

`apply`'s gate is unit-tested (including that it never reads stdin on a TTY); it could not be driven end-to-end here because `PlanApply` lists releases from Docker first.

## Follow-ups (not in this PR)

- **swarmcli-be**: stamp `charts.engineVersion` from `.swarmcli-ref` in its goreleaser — without it BE reports an unstamped engine and warns instead of refusing. Needs this PR merged first (deploy order).
- **swarmcli-charts**: scaffold `swarmcliVersion` in `new-chart.sh`; document the field.
- Deferred by decision: index-aware version resolution, `.Capabilities` in templates, `charts lint`.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -34,6 +34,7 @@ builds:
       - -X main.version={{.Version}}
       - -X main.commit={{.Commit}}
       - -X main.date={{.Date}}
+      - -X swarmcli/charts.engineVersion={{.Version}}
     main: .
 
   - id: osx
@@ -49,6 +50,7 @@ builds:
       - -X main.version={{.Version}}
       - -X main.commit={{.Commit}}
       - -X main.date={{.Date}}
+      - -X swarmcli/charts.engineVersion={{.Version}}
     main: .
 
   - id: windows
@@ -65,6 +67,7 @@ builds:
       - -X main.version={{.Version}}
       - -X main.commit={{.Commit}}
       - -X main.date={{.Date}}
+      - -X swarmcli/charts.engineVersion={{.Version}}
     main: .
 
   - id: freebsd
@@ -80,6 +83,7 @@ builds:
       - -X main.version={{.Version}}
       - -X main.commit={{.Commit}}
       - -X main.date={{.Date}}
+      - -X swarmcli/charts.engineVersion={{.Version}}
     main: .
 
 archives:
--- a/README.md
+++ b/README.md
@@ -105,6 +105,11 @@ deployed state is reproducible and an automated updater (e.g. Renovate) has
 something concrete to bump. It never removes a release the file does not mention —
 it reports those instead. See [charts/README.md](charts/README.md#declarative-releases-gitops).
 
+A chart can declare the swarmcli it needs (`swarmcliVersion: ">= 1.13.0"` in
+`Chart.yaml`). Installing it on an older build fails naming the version to
+upgrade to, instead of breaking somewhere inside the template; `--skip-compat-check`
+proceeds anyway. See [charts/README.md](charts/README.md#declaring-the-swarmcli-a-chart-needs).
+
 Run `swarmcli charts --help` for the full command and option reference.
 
 ## Business Edition
--- a/charts/README.md
+++ b/charts/README.md
@@ -131,7 +131,7 @@ version is published, with the chart's release notes attached. Merge it, and
 
 ```
 mychart/
-├── Chart.yaml          # name, version, appVersion, description, maintainers
+├── Chart.yaml          # apiVersion, name, version, appVersion, swarmcliVersion, …
 ├── values.yaml         # default values
 ├── values.schema.json  # optional JSON Schema validated before render
 ├── README.md
@@ -141,6 +141,35 @@ mychart/
     ├── secrets.yaml
     └── volumes.yaml
 ```
+
+`apiVersion` is `v1` (the only format this build reads; absent means `v1`).
+
+### Declaring the swarmcli a chart needs
+
+A chart may state the chart engine it requires, as a SemVer constraint:
+
+```yaml
+# Chart.yaml
+swarmcliVersion: ">= 1.13.0"
+```
+
+`install`, `upgrade` and `apply` refuse a chart this build is too old for, so the
+failure names the version to upgrade to rather than surfacing as whatever error
+the missing feature happens to produce (`function "toYamlPretty" not defined`).
+`install` and `upgrade` ask before refusing when run interactively; `apply` never
+does — it is meant to run unattended. `template`, `diff` and `show` only warn:
+they change nothing, and `show` is how you find out what a chart wants. Pass
+`--skip-compat-check` to proceed anyway.
+
+The constraint is checked against the **chart engine's** version, which is not
+necessarily the version the binary reports for itself — a binary embedding this
+package carries whichever engine it pinned. A build that reports no engine
+version (any `go build`) warns rather than refusing: not knowing is not evidence
+of incompatibility. Charts declaring nothing are unaffected.
+
+Note that only builds carrying this check can honour it: `Chart.yaml` is parsed
+leniently, so an older swarmcli silently ignores `swarmcliVersion` — the same
+bootstrapping limit Helm's own `apiVersion` gate has.
 
 Templates are rendered with Go `text/template` + [Sprig], exposing:
 
--- a/charts/apply.go
+++ b/charts/apply.go
@@ -35,6 +35,12 @@ type ReleasePlan struct {
 	Requirements *Requirements
 	// CurrentManifest is the deployed manifest, for diffing. Empty for an install.
 	CurrentManifest string
+	// Compat is the chart's engine requirement checked against this build.
+	// Planning records it but never acts on it: apply's contract is to plan
+	// every release before converging any, so the whole plan is gated at once
+	// by the caller — which is also the layer that knows whether blocking is
+	// appropriate for the verb being run.
+	Compat CompatFinding
 }
 
 // Plan is what apply would do to the whole swarm.
@@ -105,6 +111,7 @@ func (e *Engine) planRelease(rf *ReleaseFile, spec ReleaseSpec, src ChartSource,
 	if err != nil {
 		return ReleasePlan{}, fmt.Errorf("%s: release %q: %w", rf.Path, spec.Name, err)
 	}
+	compat := CheckCompat(ch.Metadata)
 
 	files, err := readFiles(rf.ValuesPaths(spec))
 	if err != nil {
@@ -126,13 +133,16 @@ func (e *Engine) planRelease(rf *ReleaseFile, spec ReleaseSpec, src ChartSource,
 		Release: ReleaseMeta{Name: spec.Name, Namespace: spec.Name, Revision: 1},
 		Chart:   ChartMeta{Name: rc.Name, Version: rc.Version, AppVersion: rc.AppVersion},
 	}
+	// A chart needing a newer engine usually fails here rather than reaching the
+	// caller's gate, so name the requirement alongside whatever the missing
+	// feature produced (compatHint is empty unless the chart is incompatible).
 	manifest, err := Render(ch, rctx)
 	if err != nil {
-		return ReleasePlan{}, fmt.Errorf("%s: release %q: %w", rf.Path, spec.Name, err)
+		return ReleasePlan{}, fmt.Errorf("%s: release %q: %w%s", rf.Path, spec.Name, err, compatHint(compat))
 	}
 	req, err := RenderRequirements(ch, rctx)
 	if err != nil {
-		return ReleasePlan{}, fmt.Errorf("%s: release %q: %w", rf.Path, spec.Name, err)
+		return ReleasePlan{}, fmt.Errorf("%s: release %q: %w%s", rf.Path, spec.Name, err, compatHint(compat))
 	}
 
 	rp := ReleasePlan{
@@ -143,6 +153,7 @@ func (e *Engine) planRelease(rf *ReleaseFile, spec ReleaseSpec, src ChartSource,
 		Values:       values,
 		Manifest:     manifest,
 		Requirements: req,
+		Compat:       compat,
 	}
 
 	cur, ok := deployed[spec.Name]
--- a/charts/apply_test.go
+++ b/charts/apply_test.go
@@ -67,6 +67,60 @@ releases:
     version: "0.1.0"
 `
 
+// Planning records a chart's engine requirement but must not act on it: apply
+// plans every release before converging any, so the whole plan is gated at once
+// by the caller.
+func TestPlanApplyRecordsCompatFinding(t *testing.T) {
+	withEngineVersion(t, "1.12.0")
+	e, _, src, rf := applyEnv(t, oneRelease)
+
+	ch := demoChart(t, "0.1.0")
+	ch.Metadata.SwarmcliVersion = ">= 1.13.0"
+	src.charts["swarmcli-charts/demo@0.1.0"] = ch
+
+	plan, err := e.PlanApply(context.Background(), rf, src)
+	require.NoError(t, err, "an incompatible chart must still plan; refusing is the caller's call")
+	require.Len(t, plan.Releases, 1)
+	require.Equal(t, ActionInstall, plan.Releases[0].Action)
+	require.Equal(t, CompatIncompatible, plan.Releases[0].Compat.Status)
+	require.Equal(t, ">= 1.13.0", plan.Releases[0].Compat.Required)
+}
+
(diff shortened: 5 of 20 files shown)


eldara-cruncher:
## Manual verification guide (this PR + #462 + swarmcli-charts#85)

Runnable **now, from the PR branches** — you simulate the v1.13.0 release by stamping the version with `-ldflags`, which is all the real build does. Every command below was run verbatim; the three items that genuinely need a real tag are isolated in Part E. The durable, chart-author-facing version of the floor testing lives in swarmcli-charts `docs/e2e-testing.md`; this comment is the reviewer-facing, branch-specific procedure and is expected to age out after merge.

---

# Manual testing — chart ↔ swarmcli compatibility

Covers three PRs: swarmcli **#461** (the `swarmcliVersion` field + gate),
swarmcli **#462** (`charts lint`), swarmcli-charts **#85** (declared floors +
`floor-check.sh`).

## When to run what

Almost everything below is testable **now, from the PR branches** — you fake the
release version with `-ldflags`, which is all the real build does. That is the
point of running it: it validates the PRs *before* merge.

Only these strictly need a real `v1.13.0` tag, and are called out inline as
**[post-release]**:
- the *shipped* binary is stamped (you can still dry-run this now with a
  goreleaser snapshot);
- `zammad`'s `>= 1.13.0` floor stops being "skipped, unverified" in charts CI;
- BE stamps `engineVersion` from `.swarmcli-ref`.

No Docker is needed for the gate, lint, render, or floor checks — they stop
before any daemon call. A live single-node swarm is only needed where a step says
**[needs swarm]** (an install that actually deploys, and `apply`).

---

## Setup

```bash
cd /claude-go/swarmcli-compat        # branch: feat/charts-lint (contains #461 + #462)

# The "new" binary — simulate the v1.13.0 release that ships the feature.
go build -ldflags "-X main.version=1.13.0 -X swarmcli/charts.engineVersion=1.13.0" -o /tmp/swc-new .

# An "old" released binary that predates the feature (real tag, already exists).
# --no-hardlinks: /tmp and /claude-go may be different filesystems.
git clone -q --local --no-hardlinks --branch v1.12.0 /claude-go/swarmcli /tmp/ce-old && \
  ( cd /tmp/ce-old && go build -o /tmp/swc-old . )

# A dev binary with no version stamped at all.
go build -o /tmp/swc-dev .

CHARTS=/claude-go/charts-floors/charts   # branch: feat/swarmcli-version-floors (#85)

# setfloor <chart-dir> <constraint|-> : set (or, with "-", remove) the chart's
# floor. The #85 charts already declare swarmcliVersion, so strip-then-append —
# a bare append would make a duplicate YAML key.
setfloor() {
  sed -i '/^swarmcliVersion:/d' "$1/Chart.yaml"
  [ "$2" = "-" ] || printf '\nswarmcliVersion: "%s"\n' "$2" >> "$1/Chart.yaml"
}

# A throwaway chart that declares a floor higher than the new binary provides.
rm -rf /tmp/need-next && cp -r "$CHARTS/traefik" /tmp/need-next
setfloor /tmp/need-next ">= 99.0.0"
```

---

## Part A — the compatibility gate (#461)

### A1. The bootstrap limitation — an old binary ignores the field
This is the most important and least intuitive property: the swarmcli that is
*too old* is the one that cannot warn about itself.

```bash
/tmp/swc-old charts template t /tmp/need-next  >/dev/null 2>/tmp/old.err; echo "old: exit=$? warn=[$(cat /tmp/old.err)]"
/tmp/swc-new charts template t /tmp/need-next  >/dev/null 2>/tmp/new.err; echo "new: exit=$? warn=[$(cat /tmp/new.err)]"
```
Expect: **old** → exit 0, warn empty (field silently ignored). **new** → exit 0,
warn = `chart traefik 0.1.0 requires swarmcli >= 99.0.0; this build provides 1.13.0`.

### A2. Satisfied constraint is silent
```bash
rm -rf /tmp/ok-chart && cp -r "$CHARTS/traefik" /tmp/ok-chart && setfloor /tmp/ok-chart ">= 1.0.0"
/tmp/swc-new charts template t /tmp/ok-chart >/dev/null 2>/tmp/ok.err; echo "exit=$? warn=[$(cat /tmp/ok.err)]"
```
Expect: exit 0, warn empty.

### A3. install, non-interactive → aborts (this is the "throw")
```bash
/tmp/swc-new charts install t /tmp/need-next < /dev/null; echo "exit=$?"
```
Expect: `Error: chart traefik 0.1.0 requires swarmcli >= 99.0.0; this build
provides 1.13.0` + upgrade hint, **exit 1**, and it never reaches Docker.

### A4. install, interactive → prompts, defaults to No
Run in a **real terminal** (not piped):
```bash
/tmp/swc-new charts install t /tmp/need-next
```
Expect the prompt `Continue anyway? [y/N]`. Pressing **Enter** aborts (exit 1);
**y** proceeds past the gate (then needs a swarm to actually deploy).

### A5. --skip-compat-check forces through
```bash
/tmp/swc-new charts install t /tmp/need-next --skip-compat-check < /dev/null
```
Expect: `... continuing anyway (--skip-compat-check)` on stderr, then it proceeds
(reaching Docker).

### A6. apply never prompts **[needs swarm]**
`apply` refuses non-interactively and, unlike install, does not prompt even on a
terminal. With a release file referencing an incompatible chart:
```bash
/tmp/swc-new charts apply -f your-release.yaml        # exit 1, no prompt, in a terminal too
/tmp/swc-new charts apply -f your-release.yaml --dry-run   # warns only, exit 0
```

### A7. Dev build (no stamp) never blocks
```bash
/tmp/swc-dev charts install t /tmp/need-next < /dev/null
```
Expect: `... this build reports no chart-engine version; compatibility was not
checked`, then it proceeds. Not knowing the version must never block.

### A8. apiVersion is now validated
```bash
cp -r "$CHARTS/traefik" /tmp/bad-api && printf '\napiVersion: v99\n' >> /tmp/bad-api/Chart.yaml
/tmp/swc-new charts show chart /tmp/bad-api; echo "exit=$?"
```
Expect: `Error: Chart.yaml: unsupported apiVersion "v99" (expected "v1")`, exit 1.
`v1` and absent both load fine — confirm the real charts still load:
```bash
for c in "$CHARTS"/*/; do /tmp/swc-new charts show chart "$c" >/dev/null || echo "FAIL $c"; done
```
Expect: no output (all 12 load).

---

## Part B — `charts lint` (#462)

### B1. A clean chart
```bash
/tmp/swc-new charts lint /tmp/ok-chart; echo "exit=$?"
```
Expect: `traefik 0.1.0: ok`, exit 0.

### B2. A chart with no declared floor → warning, not failure
```bash
rm -rf /tmp/no-floor && cp -r "$CHARTS/traefik" /tmp/no-floor && setfloor /tmp/no-floor -
/tmp/swc-new charts lint /tmp/no-floor; echo "exit=$?"
```
Expect: `warning: Chart.yaml declares no swarmcliVersion ...` then `... : ok`,
exit 0.

### B3. --for-version below the declared floor → error
```bash
/tmp/swc-new charts lint "$CHARTS/zammad" --for-version 1.12.0; echo "exit=$?"
```
Expect: `error: Chart.yaml requires swarmcli >= 1.13.0, which 1.12.0 does not
satisfy`, exit 1. Above the floor (`--for-version 1.13.0`) → exit 0.

### B4. Required-input chart needs its values (the renovate case)
```bash
/tmp/swc-new charts lint "$CHARTS/renovate"; echo "bare exit=$?"
/tmp/swc-new charts lint "$CHARTS/renovate" -f "$CHARTS/renovate/ci/default-values.yaml"; echo "-f exit=$?"
```
Expect: bare → `error: render ... error calling fail: set repositories ...`,
exit 1 (correct — no input). With `-f` → `renovate 0.1.0: ok`, exit 0.

### B5. A broken template
```bash
cp -r "$CHARTS/traefik" /tmp/broken && echo '{{ toYamlPretty .Values }}' > /tmp/broken/templates/stack.yaml.tmpl
/tmp/swc-new charts lint /tmp/broken; echo "exit=$?"
```
Expect: `error: render ... function "toYamlPretty" not defined`, exit 1.

---

## Part C — declared floors + CI gate (#85)

Run from the charts worktree:
```bash
cd /claude-go/charts-floors
```

### C1. floor-check verifies real floors and skips the unreleased one loudly
```bash
scripts/floor-check.sh /tmp/floorbin; echo "exit=$?"
```
Expect: 11 charts `rendering with real swarmcli v1.11.0` → `OK`; a
`zammad ... v1.13.0 is not released — floor NOT verified` line with a
`::warning::` annotation; `All declared floors hold.`; **exit 0** (green on main).

### C2. The gate actually catches a lie
Make zammad claim a *released* version it does not run on:
```bash
sed -i 's/>= 1.13.0/>= 1.12.0/' charts/zammad/Chart.yaml
scripts/floor-check.sh /tmp/floorbin; echo "exit=$?"
git checkout charts/zammad/Chart.yaml     # restore
```
Expect: `FAIL [...] — chart does not run on the v1.12.0 it declares`, **exit 1**.

### C3. test-charts.sh no longer swallows the warning
```bash
cp charts/whoami/Chart.yaml /tmp/w.bak
sed -i 's/>= 1.11.0/>= 99.0.0/' charts/whoami/Chart.yaml
SWARMCLI=/tmp/swc-new scripts/test-charts.sh 2>&1 | grep -A2 'whoami \[default\]'
cp /tmp/w.bak charts/whoami/Chart.yaml     # restore
```
Expect a `NOTE: renderer warnings` block quoting the compat message (before the
fix this printed nothing).

---

## Part D — cross-check: #462 lint agrees with #85 floor-check

`lint --for-version` reasons from the *declared* floor; `floor-check` proves it
with a *real* binary. They must agree.
```bash
for v in 1.10.0 1.11.0 1.12.0 1.13.0; do
  ok=0; for d in "$CHARTS"/*/; do
    /tmp/swc-new charts lint "$d" -f "$d/ci/default-values.yaml" --for-version "$v" 2>&1 \
      | grep -qi "does not satisfy" || ok=$((ok+1))
  done
  echo "$v: $ok/12 charts' floors admit it"
done
```
Expect: `1.10.0: 0/12`, `1.11.0: 11/12`, `1.12.0: 11/12`, `1.13.0: 12/12` — the
same shape floor-check produces (zammad the lone holdout at 1.11/1.12).

---

## Part E — [post-release] after v1.13.0 is tagged

### E1. The shipped binary is stamped (dry-run now, real after tag)
```bash
cd /claude-go/swarmcli-compat && goreleaser build --snapshot --clean --single-target -o /tmp/swc-rel
/tmp/swc-rel charts install t /tmp/need-next < /dev/null   # must gate, not warn-as-unstamped
```
The gate firing (not the A7 "no chart-engine version" message) proves the ldflag
reached the real build.

### E2. zammad's floor auto-verifies
Once `v1.13.0` exists, re-run `scripts/floor-check.sh`: zammad moves from
"NOT verified" to `rendering with real swarmcli v1.13.0 → OK`, with no code change.

### E3. BE stamps from .swarmcli-ref
After the BE goreleaser change and a BE build:
```bash
cd /claude-go/swarmcli-be
go build -ldflags "-X main.version=1.13.1 -X swarmcli/charts.engineVersion=$(cat .swarmcli-ref)" -o /tmp/swc-be .
/tmp/swc-be charts install t /tmp/need-next < /dev/null
```
Expect the two-version phrasing: `... this build provides chart engine <ce> (swarmcli 1.13.1)`.

---

## Cleanup
```bash
rm -rf /tmp/swc-new /tmp/swc-old /tmp/swc-dev /tmp/ce-old /tmp/swc-be /tmp/swc-rel \
       /tmp/need-next /tmp/ok-chart /tmp/no-floor /tmp/bad-api /tmp/broken /tmp/floorbin \
       /tmp/*.err /tmp/w.bak
cd /claude-go/charts-floors && git checkout . && rm -rf .rendered .floor-bin .swarmcli-bin
```

PR #467, opened 2026-07-18
fix(charts): suggest `charts repo update` on digest mismatch (#456)

### LINKED ISSUE #456: [charts] Change error message
```
$ ./swarmcli-be charts show values localrepo/zammad >
 zammad.yaml
Error: chart "zammad" version 0.1.0: digest mismatch (index says b3a831e8b01eb651193a1b85f96b25ad0c96fcc509889571e4caa7d80789e2cf, download is 23c44ebba4b6e7e50ec281d751c8e8812c863da12fda689fcdcae41850611489); the repository index and the chart archive disagree — refusing to install
```

Would be nice to suggest to run:
```
$ swarmcli charts repo update
```

eldara-cruncher:
Closes #456.

## What

When a chart archive's SHA-256 disagrees with the digest cached in the local repository index, `verifyDigest` aborts with a bare "digest mismatch … refusing to install". This appends a one-line hint pointing at the remedy:

```
Error: chart "zammad" version 0.1.0: digest mismatch (index says b3a8…, download is 23c4…); the repository index and the chart archive disagree — refusing to install
hint: the cached repository index may be stale (e.g. the chart was rebuilt since you last refreshed); run `swarmcli charts repo update` to update it
```

## Why

The overwhelmingly common cause of this mismatch is a chart that was rebuilt and republished after the index was last cached. `charts repo update` re-fetches `index.yaml`, so the stored digest matches the current archive and the pull succeeds.

## Notes

- The hint is worded "may be stale" rather than instructing the user to bypass anything. The digest check is an integrity guard; `charts repo update` only re-fetches the index over HTTPS and never skips verification, so the hint is safe even if the mismatch were a genuine integrity problem.
- Mirrors the existing `githubPagesHint` pattern in the same file; hint scoped to the mismatch branch only (not the no-digest or unsupported-algorithm paths).
- `swarmcli` is hardcoded, consistent with the existing `swarmcli charts repo add` hint in `repo list`. Both the CE (`swarmcli`) and BE (`swarmcli-be`) binaries share this code.

### DIFF
--- a/charts/repo.go
+++ b/charts/repo.go
@@ -412,6 +412,12 @@ func (s *RepoStore) Pull(entry IndexEntry, baseURL string) (*Chart, error) {
 // including hand-written and older ones — that does not publish digests.
 var errNoDigest = errors.New("no digest in index entry")
 
+// staleIndexHint is appended to a digest-mismatch error: the common cause is a
+// chart rebuilt and republished after the local index was cached, so refreshing
+// the index makes the two agree again.
+const staleIndexHint = "\nhint: the cached repository index may be stale (e.g. the chart was " +
+	"rebuilt since you last refreshed); run `swarmcli charts repo update` to update it"
+
 // verifyDigest checks a downloaded chart archive against the digest published in
 // the repository index. The index is fetched over HTTPS from the repository, but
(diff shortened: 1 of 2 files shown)


mosonyi (approved):
LGTM

Issue #480, opened 2026-07-21
Services view shows scheduled replicas, not running ones (diverges from docker service ls)


eldara-cruncher:
Split out of #473, which fixes the same root cause for the chart-convergence path only.

`ServiceEntry.ReplicasOnNode` is what the services view renders as the `REPLICAS` column, and it is computed by `docker/service.go`:

```go
func countTasksForNode(serviceID, nodeID string, snap *SwarmSnapshot) int {
	for _, t := range snap.Tasks {
		if t.ServiceID != serviceID { continue }
		if t.DesiredState != swarm.TaskStateRunning { continue }   // DESIRED, not actual
		if nodeID == "" || t.NodeID == nodeID { count++ }
	}
}
```

It counts tasks the orchestrator *intends* to run. Tasks are created with `DesiredState=RUNNING` and only then walk `new → pending → assigned → preparing → starting → running`, so a task that has never started is counted as if it were up.

Verified with a constructed snapshot (task list is fetched unfiltered):

```
countTasksForNode = 3   (actually running = 1)
   task 1: Status=running
   task 2: Status=pending
   task 3: Status=starting
```

## Impact

`docker service ls` shows `RunningTasks/DesiredTasks` — actually running. The services view shows scheduled/desired. So during a rollout, a node outage, or an image pull, the TUI reads `3/3` where `docker service ls` reads `1/3`. It is most misleading exactly when the operator most needs it: a service whose tasks are stuck `pending` on an unsatisfiable placement constraint displays as fully converged forever.

## Why this is separate from #473

#473 fixes the chart-convergence path, and does so by adding a correct count alongside rather than changing this one — because `ReplicasOnNode` backs the TUI and changing its meaning in place is a user-visible change that deserves its own review.

## Suggested fix

Count `Status.State == swarm.TaskStateRunning` instead, and exclude tasks on nodes that are not `Ready`/`Active` from the denominator (a task pinned to a down node never converges). #473 grows a helper doing exactly this for the convergence path; the two should share it rather than diverge.

Worth deciding deliberately: whether `REPLICAS` should also require tasks to be *up-to-date* with the current service spec. `docker service ls` does not, which is why a mid-rollout service can read `3/3` there too.

PR #523, opened 2026-07-30
fix(services): show how far a rollout got, and stop hiding scheduled tasks

eldara-cruncher:
Reported from a live cluster: a service row read `2/2` while the task table expanded under it showed **three** tasks with `DESIRED STATE = running`.

```
swarmcli-frontend_frontend   2/2   updating  2/2 healthy  replicated  …/frontend:v1.3.23
   NAME                    NODE     DESIRED STATE  CURRENT STATE
   swarmcli-frontend_fro…  node-29  running        preparing 26 seconds ago
   swarmcli-frontend_fro…  node-02  running        running 36 seconds ago
   swarmcli-frontend_fro…  node-03  running        running 2 weeks ago
```

Every number there is right, and that is the problem. Two separate defects, both from b60d2d2 (#489), plus a table that hides what would have made the row legible.

## What is actually happening

`REPLICAS` is `running / declared replicas`. The denominator is literally `*Spec.Mode.Replicated.Replicas`, so the service targets **2** replicas; the three rows are two slots' worth, one of them mid-replacement. Two containers are up, so the numerator is 2.

That matches `docker service ls` exactly — checked against the source rather than the docs: swarmkit's `ListServiceStatuses` counts `RunningTasks` as any task with `Status.State == Running`, with no desired-state and no active-node filter, and `DesiredTasks` as `replicas` ([`controlapi/service.go`](https://github.com/moby/swarmkit/blob/v2.1.2/manager/controlapi/service.go#L1083-L1124)). `docker service ls` prints `2/2` here too.

So under a **start-first** rollout both containers counted are the *outgoing* generation. The row reads fully converged while nothing has landed — and node-29 has a task that shut down 32s ago and another preparing 26s ago, i.e. a retry loop. A rollout wedged on an image pull looks converged for as long as it stays wedged. This is the case #480 flagged and deferred:

> Worth deciding deliberately: whether `REPLICAS` should also require tasks to be *up-to-date* with the current service spec. `docker service ls` does not, which is why a mid-rollout service can read `3/3` there too.

## 1. STATUS says how far the rollout got

`REPLICAS` is unchanged — the parity #489 established is the right answer to "how many replicas are serving". The other question, how many of those are the generation being rolled out, gets its own count, and `STATUS` carries it, since `updating` is what raises the question and the ratio beside it cannot answer it:

```
before   swarmcli-frontend_frontend  2/2  updating              2/2 healthy
after    swarmcli-frontend_frontend  2/2  updating · 1/2 new    2/2 healthy
```

It appears only while a rollout is in flight — `updating`, `paused`, or either rollback state — and only while it is short of target. A settled service renders exactly as before, and a pull in flight still owns the cell.

**How the current generation is identified.** The newest task per slot, by `CreatedAt`: swarm creates a slot's replacement after the task it supersedes, so creation order *is* generation order. `DesiredState` cannot draw this line — under start-first the outgoing task keeps `DesiredState=running` until its replacement is up, which is precisely the window that matters. Comparing task images against the service's `TaskTemplate` was rejected: digest resolution makes it fragile, and a force-update or env-only change leaves the images identical, so the counter would report converged immediately.

This is deliberately **not** `taskutil.LatestTasksByServiceKey`. That helper prefers a task that wants to be running over a newer terminal one, because it picks a task worth surfacing an *error* from; on a slot whose replacement was rejected that picks the outgoing task and would report the old generation as current. Two different rules, kept separate rather than one growing a mode flag.

The `--wait` path already reached the same conclusion independently — `charts/release.go:981` returns `PhaseProgressing` for `updating` with the comment "the count may still be the outgoing generation" — so this agrees with the codebase's settled position rather than inventing one.

## 2. The expanded table names the replica and the generation

The counter above says the rollout has not landed. It cannot say why the table shows three rows against a target of two, and that is not derivable from it: `2/2` plus `1/2 new` is equally consistent with two task records (swarm has not created the replacement yet) and three.

The slot answers it, and the table was throwing it away. Every task in this expansion belongs to the service on the row directly above, so `NAME` spent 22 columns re-printing that name once per row and truncated off the `.1`/`.2` suffix — the only part that differs. `swarmcli-frontend_frontend.1` and `…frontend.2` both rendered as `swarmcli-frontend_fro…`.

```
before   swarmcli-frontend_fro…  node-29  running  preparing 26 seconds ago
         swarmcli-frontend_fro…  node-02  running  running 36 seconds ago
         swarmcli-frontend_fro…  node-03  running  running 2 weeks ago

after    REPLICA  IMAGE             NODE     DESIRED STATE  CURRENT STATE
         1        frontend:v1.3.23  node-29  running        preparing 26 seconds ago
         2        frontend:v1.3.23  node-02  running        running 36 seconds ago
         1        frontend:v1.3.22  node-03  running        running 2 weeks ago
```

Two rows are replica 1, so three tasks against a target of two stops reading as a contradiction. `IMAGE` names which one is outgoing and follows the `HEALTH`/`PORTS` precedent of appearing only when it carries information — the replicas disagreeing about the image *is* the rollout, and in a settled service the column would only repeat the service row above. A settled service therefore renders **narrower** than before.

The registry and repository path are dropped (`frontend:v1.3.23`, not the full reference), since the service row already carries it and the tag is what differs. A force-update or env-only change leaves both generations on the same tag, so `IMAGE` stays hidden there while `REPLICA` still separates them. A global service's tasks carry no slot and render `—`; `NODE` is the discriminator there.

Scoped to the services-view expansion. The stacks-view expansion spans multiple services, where `NAME` is doing real work, and is untouched. `formatTaskRow` took nine positional parameters and would have taken twelve, so its cells and its column flags are structs now.

## 3. The node view stops hiding scheduled tasks

`LoadNodeServices` decided row *visibility* from the running-only count:

```go
if onNode == 0 { continue }
```

Since #489 `onNode` counts only tasks that have actually started, so a node's services view silently drops any service whose task there is preparing, pending on an unsatisfiable constraint, or rejected — the service an operator opened that node to diagnose. On the reported cluster, node-29's view does not list `swarmcli-frontend_frontend` at all right now.

Visibility now asks whether swarm *intends* a task on the node. Such a row renders `0/N`, which is the informative answer.

## Missed call sites, swept here rather than deferred

All belong to the #480 defect:

- `refreshServiceErrorsFromSnapshot` kept its own inline desired count using `len(snap.Nodes)` for global services — the exact lie #489 removed from `getServiceStackAndDesired` but missed here, so a fully healthy global service was flagged as erroring for as long as one node stayed drained or down. Both now go through an exported `(*SwarmSnapshot).DesiredReplicas`, deleting the duplicated mode switch.
- Four comments still described the pre-#489 `DesiredState` semantics of the `REPLICAS` string: `docker/convergence.go:14`, `charts/release.go:46`, `charts/release.go:986`, `charts/backend_docker.go:165`. Their conclusion survives — the convergence path must not derive parity from that string — but for the opposite reason: it now counts every generation on any node, rather than counting intent.

## Not in scope

**`REPLICAS` semantics** stay as #489 left them.

**BE needs no change.** Its `2/2 healthy` is computed over a different task set (it additionally requires `DesiredState == running`) and will legitimately keep disagreeing with `REPLICAS` mid-rollout. BE builds and tests green against this tree.

**Relocating `views/taskutil` into `docker`** is the tidier structure — it is snapshot logic, not view logic, and both slot-latest rules could then share one key function. Held off deliberately: `swarmcli-be/bootstrap/health.go:8` imports it, so the move removes an exported package from this module's API. That is a `C1-breaking-change`, which forces a major tag at release, and rc5 is mid-flight. Worth a deliberate window with the BE import updated alongside, not a rider on a bugfix.

## Tests

`docker/service_test.go` — the reported scenario as a fixture (`2/2` with `UpToDate == 1`); the overlap moment right after, when the incoming task starts and `REPLICAS` reads `3/2` while the generation count stays bounded by the slots; steady state; a global service keyed by node; an unassigned task not colliding on an empty key; `isRollingOut` across all six update states plus nil; `hasIntendedTaskOnNode` for preparing / terminal-only / other-service nodes.

`views/services/columns_test.go` — the STATUS counter renders while rolling out and short, disappears at parity, stays absent outside a rollout, loses to `PullProgress`, and is skipped when there is no target; the expansion shows `REPLICA` and drops the repeated service name, shows `IMAGE` only when generations disagree, renders `—` for a global service's slot, and `shortImage` keeps the tag while dropping a registry path (including one with a port).

`gofmt`, `go vet ./...`, `go test -race -count=1 ./...`, `golangci-lint run ./... --build-tags=integration` → 0 issues.

## Verification on a real cluster

Not verified against live Swarm — no cluster access from here. Worth checking manually:

1. The reported service's row shows `updating · 0/2 new` or `· 1/2 new` while node-29 keeps retrying, and drops back to a bare `updating` → `updated` as replicas land.
2. Expanding it shows `REPLICA 1` twice — the outgoing `v1.3.22` and the incoming `v1.3.23` — so three rows against a target of two reads correctly.
3. A settled service's expansion shows no `IMAGE` column and is no wider than before.
4. Node-29's services view lists `swarmcli-frontend_frontend` again, at `0/2`.
5. `docker service inspect swarmcli-frontend_frontend --format '{{.Spec.Mode.Replicated.Replicas}}'` prints `2` — the premise of the diagnosis above. If it prints `3`, the denominator is a separate bug and this PR does not address it.

Refs #480.

### DIFF
--- a/charts/backend_docker.go
+++ b/charts/backend_docker.go
@@ -167,10 +167,11 @@ func (b *dockerBackend) StackServices(ctx context.Context, name string) []Servic
 // swarms is not forced through the ambient one.
 func ServiceStatesFrom(snap *docker.SwarmSnapshot, name string) []ServiceState {
 	entries := snap.StackServices(name)
-	// Convergence facts come from a separate loader because ServiceEntry counts
-	// replicas by desired state, which is right for the services view but wrong
-	// for deciding a rollout finished (issues #473, #480). Both now read the
-	// same snapshot, so the display and the decision cannot disagree.
+	// Convergence facts come from a separate loader because ServiceEntry mirrors
+	// `docker service ls` — counting every generation, on any node — which is
+	// right for the services view but wrong for deciding a rollout finished
+	// (issues #473, #480). Both now read the same snapshot, so the display and
+	// the decision cannot disagree.
 	conv := make(map[string]docker.ServiceConvergence, len(entries))
 	for _, c := range snap.StackConvergence(name) {
 		conv[c.Name] = c
--- a/charts/release.go
+++ b/charts/release.go
@@ -47,9 +47,10 @@ type ServiceState struct {
 	Status   string
 
 	// Running counts tasks that are actually running on an active node.
-	// Deliberately not derived from Replicas: that string counts tasks by
-	// DESIRED state, so it reaches its target the moment Swarm schedules the
-	// tasks rather than when they are up (see issue #480).
+	// Deliberately not derived from Replicas: that string mirrors `docker
+	// service ls` and so counts every generation, superseded tasks included, and
+	// counts them wherever they sit — a down node's stale task included (see
+	// issue #480).
 	Running int
 	// Desired is the target task count over active nodes.
 	Desired int
@@ -1178,10 +1179,10 @@ func (s ServiceState) Convergence() Convergence {
 		return Convergence{PhaseProgressing, "rolling back"}
 	}
 	// Parity is measured from the actual running count, not the Replicas display
-	// string. That string counts tasks by DESIRED state, so it reaches parity
-	// the instant swarm schedules the tasks — before any container is up — which
-	// made --wait return immediately on a fresh install, where UpdateState is
-	// empty and the in-flight arm above cannot fire either (issue #473).
+	// string. That string counts every generation on any node, so it can reach
+	// parity on tasks this package would not accept — which made --wait return
+	// early on a fresh install, where UpdateState is empty and the in-flight arm
+	// above cannot fire either (issue #473).
 	//
 	// A job's task ends Complete and is never replaced, so requiring a running
 	// task would report a step that succeeded as one that never came up. A
--- a/docker/convergence.go
+++ b/docker/convergence.go
@@ -11,9 +11,10 @@ import (
 
 // ServiceConvergence is what a caller needs to decide whether one service has
 // finished rolling out. It is deliberately separate from ServiceEntry: that
-// struct backs the services view, and its ReplicasOnNode counts tasks by
-// DESIRED state, which answers "what does the orchestrator intend" rather than
-// "what is actually up" (see issue #480).
+// struct backs the services view, where ReplicasOnNode mirrors `docker service
+// ls` and so keeps counting a superseded task while its container runs. Running
+// here drops the outgoing generation, which is the question --wait asks and the
+// services view answers with UpToDate instead (see issue #480).
 type ServiceConvergence struct {
 	Name string
 	Mode string
@@ -175,6 +176,14 @@ func isJobService(svc swarm.Service) bool {
 	default:
 		return false
 	}
+}
+
+// DesiredReplicas is the service's target task count against this snapshot: the
+// declared replicas, or for a global service one per node that can currently run
+// one. Exported for callers outside this package that would otherwise duplicate
+// the mode switch and get the global case wrong (issue #480).
+func (snap *SwarmSnapshot) DesiredReplicas(svc swarm.Service) int {
+	return desiredOverActiveNodes(svc, len(activeNodeIDs(snap)))
 }
 
 // desiredOverActiveNodes is the target task count. For a global service that is
--- a/docker/service.go
+++ b/docker/service.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"os/exec"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -352,10 +353,21 @@ type ServiceEntry struct {
 	ServiceID      string
 	ReplicasOnNode int
 	ReplicasTotal  int
-	Status         string
-	Mode           string
-	Image          string
-	Ports          string
+	// UpToDate counts the replicas running the service's *current* generation.
+	// ReplicasOnNode counts every running task, superseded ones included, so it
+	// matches `docker service ls` — which means a start-first rollout reads as
+	// fully converged while the outgoing generation is what is actually up. The
+	// two together say what one ratio cannot: how many replicas are serving, and
+	// how many of those are the version being rolled out (issue #480).
+	UpToDate int
+	// RollingOut reports a rollout in flight — updating, paused, or either
+	// rollback state. It gates the display of UpToDate: outside a rollout a
+	// replica short of the current generation is a restart, not a stale version.
+	RollingOut bool
+	Status     string
+	Mode       string
+	Image      string
+	Ports      string
 	// Health is an aggregate health summary for the service's running replicas
 	// (e.g. "2/2 healthy"); "" when unknown. The swarm API does not expose
 	// container health, so the default loaders leave it empty; it is an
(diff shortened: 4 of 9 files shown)


eldara-cruncher:
## Follow-up: the summary row alone does not explain the three task rows

Reviewing the first commit against the reported screenshot, `updating · 1/2 new` says the rollout has not landed but cannot say why the table below shows **three** rows against a target of two — and it is not derivable from it either. `2/2` plus `1/2 new` is equally consistent with two task records (swarm has not created the replacement) and three (it has, still preparing).

The slot answers it, and the table was throwing it away. Every task in this expansion belongs to the service on the row directly above, so `NAME` spent 22 columns re-printing that name once per row and truncated off the `.1`/`.2` suffix — the only part that differs. `swarmcli-frontend_frontend.1` and `…frontend.2` both rendered as `swarmcli-frontend_fro…`.

```
before   swarmcli-frontend_fro…  node-29  running  preparing 26 seconds ago
         swarmcli-frontend_fro…  node-02  running  running 36 seconds ago
         swarmcli-frontend_fro…  node-03  running  running 2 weeks ago

after    REPLICA  IMAGE             NODE     DESIRED STATE  CURRENT STATE
         1        frontend:v1.3.23  node-29  running        preparing 26 seconds ago
         2        frontend:v1.3.23  node-02  running        running 36 seconds ago
         1        frontend:v1.3.22  node-03  running        running 2 weeks ago
```

Two rows are replica 1, so three tasks against a target of two stops reading as a contradiction. `IMAGE` names which one is the outgoing generation and follows the `HEALTH`/`PORTS` precedent of appearing only when it carries information — the replicas disagreeing about the image *is* the rollout, and in a settled service the column would only repeat the service row above. A settled service therefore renders **narrower** than before.

The registry and repository path are dropped (`frontend:v1.3.23`, not the full reference) since the service row already carries it and the tag is what differs. A force-update or env-only change leaves both generations on the same tag, so `IMAGE` stays hidden there while `REPLICA` still separates them. A global service's tasks carry no slot and render `—`; `NODE` is the discriminator there.

Scoped to the services-view expansion — the stacks-view expansion spans multiple services, where `NAME` is doing real work and is untouched.

Also swept three comments in `charts/` that still described the pre-#489 `DesiredState` semantics of the `REPLICAS` string (`release.go:46`, `release.go:986`, `backend_docker.go:165`). Their conclusion survives — the convergence path must not derive parity from that string — but for the opposite reason: it now counts every generation on any node, rather than counting intent.

### Deliberately not done: relocating `views/taskutil`

`countUpToDateTasks` is a second slot-latest rule alongside `taskutil.LatestTasksByServiceKey`, and the tidier structure is to move that package into `docker` — it is snapshot logic, not view logic — so both rules share one key function. Holding off: `swarmcli-be/bootstrap/health.go:8` imports `views/taskutil`, so the move removes an exported package from this module's API. That is a `C1-breaking-change`, which forces a major tag at release, and rc5 is mid-flight. Worth doing in a deliberate window with the BE import updated alongside, not here.

The two rules genuinely differ, which is why the duplication is not simply waste: `LatestTasksByServiceKey` prefers a task that wants to be running over a newer terminal one, because it picks a task worth surfacing an *error* from. On a slot whose replacement was rejected, that picks the still-running outgoing task and would report the old generation as current.

PR #562, opened 2026-08-11
fix(ui): fullscreen uses the whole terminal

### LINKED ISSUE #560: Full screen log is not full screen
### SwarmCLI version

1.13.1

### OS / platform

Linux

### Steps to reproduce

Image

Image

### Expected behaviour

Use the whole height in full screen as well

### Actual behaviour

It is not used

eldara-cruncher:
Fixes #560.

## What was wrong

Pressing `f` left a dead band four rows deep at the bottom of the terminal. Two deductions for the frame's borders still ran in fullscreen, where `RenderViewFrame` draws no borders at all:

- `updateForResize` took two rows off the terminal height — `usableHeight = msg.Height - 2 // Just for top/bottom borders`
- every view's `FrameContent` took two more, via `ComputeFrameDimensions`' `verticalPadding`

Views with a footer lost a further row, because the fullscreen frame dropped the footer that the view had already reserved space for. Measured on a 40-row terminal, the logs view rendered 36 rows.

## What changed

The fullscreen layout now spends one row on its title line and nothing else, and `RenderViewFrame` trims or pads to exactly the height it is given **in both modes** — so no view can leave the bottom of the screen unclaimed, whatever its own sizing does. `handleViewResize` hands a fullscreen view the frame height that leaves it the whole terminal once it has made the deduction every view makes.

Three defects in the same arithmetic came with it:

- **Every resize outside the window-size handler hard-coded normal mode.** Drilling into a view, or closing an input bar, while fullscreen sized it for chrome that is not drawn — ten rows lost rather than four. They now go through `resizeToTerminal`, the one place the layout mode and the input bar are accounted for. That also fixes normal mode deducting the input bar's three rows twice, once in `updateForResize` and again in `View`.
- **`:` and `/` were invisible in fullscreen.** `View` returned before composing either bar, so the bar accepted keystrokes you could not see. Both now render in fullscreen.
- **The newest log line was never visible.** The `logs` and `inspect` viewports were sized to the whole frame while the frame drew three rows fewer, so `GotoBottom` anchored a window taller than the rows actually drawn and AutoScroll parked the newest lines in the cut-off rows. Present in normal mode too, not only fullscreen. Both viewports are now sized to the rows the frame draws, which also removes the duplicated content arithmetic from their `FrameContent`.

The fullscreen frame also renders the view's footer now (row counter, deploy progress, toasts), instead of dropping it while its rows stayed reserved.

## Verification

`go test ./...`, `go vet`, `golangci-lint run ./... --build-tags=integration` — all clean.

New guards, each confirmed to fail when the corresponding fix is reverted:

| Guard | Reverting … |
|---|---|
| `app: TestViewFillsTerminalHeight` (7 layout combinations) | the border reservation, the fullscreen trim, the double input-bar deduction |
| `app: TestFullscreenKeepsTheViewFooter` | the fullscreen footer |
| `app: TestInputBarStaysVisibleInFullscreen` | the fullscreen input bar |
| `app: TestNavigatingInFullscreenKeepsTheFullscreenBudget` | `switchToView`'s hard-coded mode |
| `app: TestHandleViewResizeFullscreenBudget` | the resize arithmetic |
| `ui: TestRenderViewFrame_OccupiesExactlyHeight` | the fullscreen trim/pad |
| `logs: TestFollowShowsTheNewestLine`, `inspect: TestFrameContentFillsTheFrame` | the viewport sizing |

`TestViewFillsTerminalHeight` counts rows that actually carry content, not just total rows — padding the reserved space with blanks looks identical to a row count and identical to the user.

Two existing assertions changed on purpose: `logs`/`inspect` `TestUpdate_WindowSizeMsg` encoded `viewport.Height == msg.Height`, which was the defect.

Not verified interactively — this was measured and asserted headlessly, not driven in a live terminal.

### DIFF
--- a/app/model.go
+++ b/app/model.go
@@ -137,7 +137,7 @@ func (m *Model) switchToView(name string, data any) tea.Cmd {
 	exitCmd := m.currentView.OnExit()
 
 	newView, loadCmd := factory(m.deps, m.viewport.Width, m.viewport.Height, data)
-	resizeCmd := handleViewResize(newView, m.viewport.Width, m.viewport.Height, false)
+	resizeCmd := handleViewResize(newView, m.viewport.Width, m.viewport.Height, m.fullscreen)
 
 	// Push current view onto stack
 	m.viewStack.Push(m.currentView)
@@ -159,7 +159,7 @@ func (m *Model) replaceView(name string, data any) tea.Cmd {
 	exitCmd := m.currentView.OnExit()
 
 	newView, loadCmd := factory(m.deps, m.viewport.Width, m.viewport.Height, data)
-	resizeCmd := handleViewResize(newView, m.viewport.Width, m.viewport.Height, false)
+	resizeCmd := handleViewResize(newView, m.viewport.Width, m.viewport.Height, m.fullscreen)
 
 	m.currentView = newView
 	m.viewStack.Reset()
--- a/app/update.go
+++ b/app/update.go
@@ -11,6 +11,7 @@ import (
 	cmdpkg "github.com/Eldara-Tech/swarmcli/commands/command"
 	"github.com/Eldara-Tech/swarmcli/docker"
 	"github.com/Eldara-Tech/swarmcli/settings"
+	"github.com/Eldara-Tech/swarmcli/ui"
 	swarmlog "github.com/Eldara-Tech/swarmcli/utils/log"
 	"github.com/Eldara-Tech/swarmcli/views/commandinput"
 	"github.com/Eldara-Tech/swarmcli/views/confirmdialog"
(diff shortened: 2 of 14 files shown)


mosonyi (changes requested):
When I see the logs and type 'f' it doesn't fill-up the whole screen until I press the 'up' key.

Image

After that it is perfect.

Image

eldara-cruncher:
Good catch — reproduced and fixed in 7d4fc8c.

The frame was the right size; the viewport inside it was not looking at the right part of the log. A resize changes how many lines fit but not the offset the viewport scrolls from, so the viewport `f` had just grown went on drawing the same lines it drew before and padded the rows it gained — the blank band in your first screenshot. `up` clamps the offset back into range (`SetYOffset` clamps to `maxYOffset`), which is why one keypress put it right, and why it stayed right afterwards.

The same stale offset hid content in the other direction, which is worth knowing since it is the keystroke right after yours: leaving fullscreen shrinks the viewport away from the bottom, so the title reads `AutoScroll: on` while the newest lines sit below the last row until the next line arrives.

Both viewports (logs and inspect) now re-anchor on a resize — to the newest line while following, and otherwise only when the offset has been left past the end, so a reader who scrolled up stays where they were.

On why the existing guards were green: the app-level test drives a stub view that rebuilds its content for whatever height it is handed, so it has no offset that can go stale. `TestFullscreenFillsTheScreenWithARealViewport` now drives the real inspect view through the `f` key instead. Each new guard was confirmed to fail with its fix reverted:

| Guard | Reverting … |
|---|---|
| `app: TestFullscreenFillsTheScreenWithARealViewport` | the inspect re-anchor (blank last row after `f`) |
| `logs: TestResizeKeepsFollowOnTheNewestLine` | the logs re-anchor, in both directions |
| `logs: TestResizeWithFollowOffClampsWithoutJumping` | the logs re-anchor (and pins that a scrolled-up reader is not dragged along) |
| `inspect: TestGrowingTheFrameLeavesNoBlankRows` | the inspect re-anchor |

`go test ./...`, `go vet`, `golangci-lint run ./... --build-tags=integration` all clean. Still measured headlessly rather than driven in a live terminal, so a second look at the same keystrokes would be welcome.

mosonyi (approved):
LGTM

PR #568, opened 2026-08-13
feat(logs): the view's options read as a status row, not a sentence

### LINKED ISSUE #561: Log view enhacement
### Problem

The log viewing options are not so visible.

### Proposed behaviour

Like in K9S

Image

### Alternatives considered

_No response_

eldara-cruncher:
## Summary

- The log view's four toggles move out of the frame title — where they were one uniform-coloured run of prose, `Service: web • AutoScroll: on • wrap: on • Filter: all nodes • Stopped: hidden`, centred in the top border and wider than the box on most terminals — and into the header row, as labelled `Label:Value` items coloured by state: green engaged, dim not, the title's scope colour for a value that is not a yes/no. The header row previously spent itself on the word "Logs".
- The title takes the k9s `ScopedTitle` form every other list view already uses: `Logs(<stack>/<service>)[<lines>]`, with the `/` filter appended as the usual `</query>` fragment. The count is post-filter, so it tracks the node, hide-stopped and `/` filters.
- `ui.ToggleRow` is the new primitive, beside the other frame primitives in `ui/`. The items sit at a fixed gap and the block is centred under the title, so the row reads as one status line: the slack is *not* handed to the gaps, which on a wide terminal drives the first and last item to opposite edges and leaves four unrelated-looking fragments 30 columns apart. It never wraps and never overhangs: as the terminal narrows the gaps close to a single space first, and only then are items dropped from the right, the cut marked with `…` rather than truncating mid-item, where a value cut to `Hid` reads as a state. The row must stay exactly one line — `views/logs/update.go` sizes the viewport to what is left after the header, so a wrapped header silently takes rows off the log content.
- The row is also the only place the toggles appear in fullscreen, where the helpbar naming the keys is not drawn. The `ctrl+f` match counter joins it as `Search:<term>(2/7)` while a search runs.
- Also fixes an over-long title breaking the frame, which the old logs title triggered on any terminal narrower than it. Both layouts assumed a title that fits: the bordered one clamps its padding at zero and drew a top border wider than the rest of the box, and the fullscreen one centres the title in a fixed width, so lipgloss wrapped it onto rows the frame had not budgeted for — 8 rows drawn for a height of 6, undoing the guarantee #560 had just established. Both now truncate through one helper.

No keys change and no behaviour changes: the same four toggles, the same helpbar. `ui.Toggle`, `ui.ToggleTone` and `ui.ToggleRow` are additions; no exported signature is changed or removed.

Rendered at 100, 70 and 40 columns:

```
┌──────────────────────────────────────── Logs(shop/web)[1432] ────────────────────────────────────────┐
│           Autoscroll:On   Wrap:Off   Node:worker-2   Stopped:Hidden   Search:timeout(2/7)            │
│ 2026-08-13T09:14:02Z  request accepted                                                               │
└──────────────────────────────────────────────────────────────────────────────────────────────────────┘

┌───────────────────────── Logs(shop/web)[1432] ─────────────────────────┐
│         Autoscroll:On Wrap:Off Node:worker-2 Stopped:Hidden …          │
│ 2026-08-13T09:14:02Z  request accepted                                 │
└────────────────────────────────────────────────────────────────────────┘

┌────────── Logs(shop/web)[1432] ──────────┐
│  Autoscroll:On Wrap:Off Node:worker-2 …  │
│ 2026-08-13T09:14:02Z  request accepted   │
└──────────────────────────────────────────┘
```

## Test Plan

- [x] `go build ./...`
- [x] `go test ./...` — full suite green
- [x] `golangci-lint run ./...` — 0 issues
- [x] `go test ./ui/... ./views/logs/...` — covers the new `ToggleRow` (centred block, uniform gaps, gaps close before any item is dropped, always one line, drops whole items when narrow, truncates a single overlong item, empty, tones distinct), the title (scope, post-filter count, app filter), the header (reflects every toggle, search state, always one row), and the frame (`RenderFramedBox` truncates an overlong title, `RenderViewFrame` keeps its height)
- [ ] Manual: open a service's logs, toggle autoscroll / wrap / node filter / hide-stopped, confirm each value flips colour in the header row
- [ ] Manual: enter fullscreen (helpbar hidden) and confirm the toggles are still readable and the frame height is exact
- [ ] Manual: on a wide terminal, confirm the row stays a centred block under the title rather than spreading to both edges
- [ ] Manual: narrow the terminal to ~70 and ~40 columns and confirm the gaps tighten first, then items drop from the right with `…` rather than truncating mid-value, and the frame border stays intact
- [ ] Manual: `ctrl+f` search, confirm `Search:<term>(n/m)` appears in the row and `n`/`N` advance it

Fixes #561

### DIFF
--- a/ui/framedbox.go
+++ b/ui/framedbox.go
@@ -19,7 +19,7 @@ func RenderViewFrame(title, header, content, footer string, width, height int, f
 		titleLine := lipgloss.NewStyle().
 			Width(width).
 			Align(lipgloss.Center).
-			Render(styleFrameTitle(title))
+			Render(fitTitle(styleFrameTitle(title), width))
 
 		rows := []string{titleLine}
 		if header != "" {
--- a/ui/framedbox_test.go
+++ b/ui/framedbox_test.go
@@ -41,6 +41,29 @@ func TestRenderViewFrame_OccupiesExactlyHeight(t *testing.T) {
 	}
 }
 
+// TestRenderViewFrame_LongTitleKeepsTheHeight — the fullscreen title line is
+// centred in a fixed width, which wraps an over-long title onto further rows
+// and pushes the frame past the height it was given.
+func TestRenderViewFrame_LongTitleKeepsTheHeight(t *testing.T) {
+	title := strings.Repeat("Logs(a-long-stack/a-long-service)", 3)
+
+	fullscreen := RenderViewFrame(title, "hdr", "body", "ftr", 40, 8, true)
+	require.Equal(t, 8, lipgloss.Height(fullscreen))
+	for i, row := range strings.Split(fullscreen, "\n") {
+		require.LessOrEqual(t, lipgloss.Width(row), 40, "fullscreen row %d", i)
+	}
+
+	// The bordered layout draws to width+4; every row of the box, the title's
+	// top border included, has to be that same width.
+	framed := RenderViewFrame(title, "hdr", "body", "ftr", 40, 8, false)
+	require.Equal(t, 8, lipgloss.Height(framed))
+	rows := strings.Split(framed, "\n")
+	bottom := lipgloss.Width(rows[len(rows)-1])
+	for i, row := range rows {
+		require.Equal(t, bottom, lipgloss.Width(row), "framed row %d", i)
+	}
+}
+
 // TestRenderViewFrame_FullscreenKeepsHeaderAndFooter — fullscreen dropped the
 // footer while the caller had already reserved its rows.
 func TestRenderViewFrame_FullscreenKeepsHeaderAndFooter(t *testing.T) {
--- a/ui/toggles.go
+++ b/ui/toggles.go
@@ -37,14 +37,21 @@ type Toggle struct {
 	Tone  ToggleTone
 }
 
-// ToggleRow renders items as "Label:Value" spread evenly across width, so a
-// view's options read at a glance instead of hiding in prose. The result is
-// always a single line no wider than width: the row is a frame header, and a
-// header that grows a second line silently shrinks the content the frame draws.
+// toggleGap is the space between items once they all fit. The slack is not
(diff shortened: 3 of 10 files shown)


mosonyi (changes requested):
Image

eldara-cruncher:
Fixed in fdd089a — you were right, and the screenshot is the case the design was worst at.

`ToggleRow` handed all the slack to the gaps so the row ended flush with both edges. That reads fine at 80 columns and falls apart at 200: the items drift into four unrelated labels 30 columns apart, with nothing tying them to the title above.

The gap is now capped at three spaces (the help bar's column gap) and the block is centred under the title. Narrowing closes the gaps to a single space before any item is dropped, so a terminal that can still hold every toggle keeps it.

```
┌──────────────────────────────────────── Logs(shop/web)[1432] ────────────────────────────────────────┐
│           Autoscroll:On   Wrap:Off   Node:worker-2   Stopped:Hidden   Search:timeout(2/7)            │
│ 2026-08-13T09:14:02Z  request accepted                                                               │
└──────────────────────────────────────────────────────────────────────────────────────────────────────┘

┌───────────────────────── Logs(shop/web)[1432] ─────────────────────────┐
│         Autoscroll:On Wrap:Off Node:worker-2 Stopped:Hidden …          │
│ 2026-08-13T09:14:02Z  request accepted                                 │
└────────────────────────────────────────────────────────────────────────┘

┌────────── Logs(shop/web)[1432] ──────────┐
│  Autoscroll:On Wrap:Off Node:worker-2 …  │
│ 2026-08-13T09:14:02Z  request accepted   │
└──────────────────────────────────────────┘
```

Two tests changed with it: the old one asserted the row was exactly as wide as the frame, and is now a centred-block assertion; a new one pins that the gaps tighten before an item is dropped. Full suite and `golangci-lint` green, and the PR description is updated to match.

mosonyi (approved):
Very nice!

PR #572, opened 2026-08-14
feat(contexts): update a context's host, keeping its TLS material

### LINKED ISSUE #570: context update
Sometimes it is required to change the host in the context.
this is the command
```
docker context update <context-name> --docker "host=tcp://<new_ip>:<port>"
```

Add this function please.

docker context update <name> --docker "host=..." overwrites the entire Docker endpoint config, not just the host field. If the context was using TLS (ca, cert, key), and you only pass host=, those TLS settings get wiped out — because update doesn't merge, it replaces the whole --docker block with whatever you provide.

This is why the IP update "worked" (connection changed) but TLS silently disappeared.

How to fix it properly

You need to pass the host and the TLS paths together, in the same command:

bash
docker context update <context-name> \
  --docker "host=ssh://user@new-ip,ca=/path/to/ca.pem,cert=/path/to/cert.pem,key=/path/to/key.pem"

If it's SSH-based (ssh://user@ip), you typically don't need TLS certs at all — SSH handles encryption/auth itself. TLS certs are only relevant for tcp://ip:2376 style endpoints (TCP + TLS).

Finding the original TLS paths

If you don't remember the original cert paths, check a still-working context for reference:

bash
docker context inspect <working-context-name>

Look for Endpoints.docker.TLSData or TLSMaterial in the output — the actual cert files live under:

~/.docker/contexts/tls/<context-hash>/
Rule of thumb

Any time you docker context update a context that has TLS materials attached, always include the full --docker string (host + ca + cert + key) in one go — never just the host — or you'll wipe the TLS config every time.

Was your affected context SSH-based or TCP+TLS? That determines whether this is even relevant for you.

eldara-cruncher:
Closes #570.

`docker context update <name> --docker "host=…"` does not merge. It replaces the
whole Docker endpoint and then resets the stored TLS material to exactly what
the command named, so passing a bare `host=` deletes a TLS context's
certificates — which is what the issue reporter hit ("the IP update worked but
TLS silently disappeared"). Reproduced against Docker before writing anything:

```
$ docker context create probe --docker "host=tcp://127.0.0.1:2376,ca=…,cert=…,key=…"
$ docker context inspect probe   # TLSMaterial: {"docker":["ca.pem","cert.pem","key.pem"]}
$ docker context update probe --docker "host=tcp://127.0.0.1:2377"
$ docker context inspect probe   # TLSMaterial: {}   ← the whole TLS dir is gone
```

The certificates a context was created from are not recorded anywhere; the
store keeps its own copies of the bytes. So the material an update re-supplies
comes from the store itself: `docker context inspect` reports `Storage.TLSPath`
and the filenames under it, and `docker context update` reads those files
before it clears the directory, which makes pointing `ca=`/`cert=`/`key=` back
at them safe. `SkipTLSVerify` rides along the same way.

## What this adds

* `docker.UpdateContextEndpoint(name, description, host)` — updates a context,
  carrying its TLS material across a host change. It replaces
  `UpdateContextWithCertFiles`, which built exactly the destructive bare-`host=`
  argv and had had no caller since #129, and `UpdateContextDescription`.
* `docker.InspectContextEndpoint` / `ContextEndpoint` — the stored endpoint,
  with the cert paths resolved into the context store.
* The contexts view's edit dialog (`e`) gains a **Host** field beside the
  description, both prefilled; `Tab`/`↑`/`↓` move between them.
* `:context update <name> --host <endpoint>` does the same from the command bar
  (`:contexts` / `:ctx` too — aliases inherit the spec). This is the first
  command in the repo to declare a flag, so `registry.FlagSpec` gets its first
  real use.
* When the endpoint of the **active** context moves, the app reconnects: the
  cached Docker client and snapshot were built for the daemon we just left, so
  the update takes the same path a context switch does.

## Two smaller fixes in the same code

* `docker context update default` can only ever fail — `default` is a reserved
  name. The dialog now says so instead of opening a form that cannot be saved,
  and the docker layer refuses it for every caller.
* Docker **ignores an empty `--description`**, so the old dialog reported
  "Updated context" while the description stayed as it was. Nothing passes an
  empty `--description` any more, and emptying the field is refused with the
  reason.

## Not included

`--description` on the command bar: the `:` parser splits on whitespace and does
not honour quotes, so a multi-word description cannot be expressed there. It
stays a dialog-only field.

## Verification

* Unit tests over the argv builder and the inspect parsing, including the two
  cases that define the feature: a host change restates ca/cert/key, and a
  description-only update passes no `--docker` at all.
* Integration tests (`integration-tests/docker/`) create a throwaway TLS
  context against a real Docker CLI and assert the stored material is
  byte-identical after a host change.
* Both were mutation-checked: reintroducing the bare-`host=` argv turns the
  unit test *and* the integration test red.

## Note for review

`ContextOps` loses two methods and gains one. Nothing in swarmcli, swarmcli-be
or swarmcli-cd referenced the removed exported functions, and no user-facing
behaviour regresses, so this is labelled `C0-breaks-nothing` — say the word if
you would rather treat an exported-symbol removal as `C1`.

### DIFF
--- a/commands/api/parse_test.go
+++ b/commands/api/parse_test.go
@@ -159,6 +159,30 @@ func TestParseInput_UnknownFlagRejected(t *testing.T) {
 	require.Contains(t, err.Error(), "unknown flag --bogus for :node")
 }
 
+// ":contexts" is the first command declaring a flag, so its spec is what lets
+// --host through strict validation and consume the following token.
+func TestParseInput_ContextsUpdateFlags(t *testing.T) {
+	cmd, a, err := ParseInput("contexts update prod --host tcp://10.0.0.7:2376")
+	require.NoError(t, err)
+	require.Equal(t, "contexts", cmd.Name())
+	require.Equal(t, []string{"update", "prod"}, a.Positionals)
+	require.Equal(t, "tcp://10.0.0.7:2376", a.Get("host"))
+
+	_, _, err = ParseInput("contexts update prod --hosts tcp://10.0.0.7:2376")
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "unknown flag --hosts for :contexts")
+}
+
+// Aliases resolve to the primary's spec, so :ctx accepts --host too.
+func TestParseInput_ContextAliasesInheritFlags(t *testing.T) {
+	for _, name := range []string{"context", "ctx"} {
+		_, a, err := ParseInput(name + " update prod --host tcp://10.0.0.7:2376")
+		require.NoError(t, err, name)
+		require.Equal(t, "tcp://10.0.0.7:2376", a.Get("host"), name)
+		require.Equal(t, []string{"update", "prod"}, a.Positionals, name)
+	}
+}
+
 func TestParseInput_PassthroughSkipsValidation(t *testing.T) {
 	// bootstrap (OSS stub) is Passthrough: --upgrade is not rejected and
 	// --help is not intercepted; everything reaches Execute.
--- a/commands/command/contexts.go
+++ b/commands/command/contexts.go
@@ -4,12 +4,26 @@
 package command
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/Eldara-Tech/swarmcli/args"
+	"github.com/Eldara-Tech/swarmcli/docker"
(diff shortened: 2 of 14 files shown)


mosonyi:
LGTM

Issue #526, opened 2026-07-31
--wait reports a release Swarm rolled back as successfully converged


eldara-cruncher:
Surfaced by a whole-repository audit of `swarmcli-cd` (Eldara-Tech/swarmcli-cd#114), but this is a CE defect that affects `--wait` for every user, independently of the controller. Verified against `moby/swarmkit@v2.1.2` in the module cache.

## What

`ServiceState.Convergence()` (`charts/release.go:972`) switches over four update states:

```go
	case "paused":            // PhaseWedged
	case "rollback_paused":   // PhaseWedged
	case "updating":          // PhaseProgressing
	case "rollback_started":  // PhaseProgressing
```

**`rollback_completed` is absent**, so it falls through to the parity + stability-window check below.

## Why that is wrong

swarmkit sets `ROLLBACK_COMPLETED` in `completeUpdate` (`manager/orchestrator/update/updater.go:639-641`) *after* `rollbackUpdate` has already restored the previous spec (`updater.go:610-616`):

```go
		service.Spec = *service.PreviousSpec
		service.SpecVersion = service.PreviousSpecVersion.Copy()
		service.PreviousSpec = nil
```

So at `rollback_completed` the tasks at parity are running the **previous** spec — exactly the state the deploy was trying to replace — and once `stabilityRemaining()` reaches zero (the 5s `defaultStabilityWindow`) `Rollup` reports `PhaseConverged`.

The codebase already knows this state exists: `docker/service.go:566` renders `swarm.UpdateStateRollbackCompleted` as `"rolled back"`. It is only the convergence switch that omits it.

## Impact

`charts/release.go:318-319`:

```go
	if opts.Wait {
		if err := e.waitReady(rel.Name, opts.Timeout); err != nil {
```

and `waitReady` (`:932`) is `Rollup(e.Backend.StackServices(release))`. `opts.Wait` comes from `f.wait` — the `--wait` flag on the chart/apply commands (`cli/apply.go:92`, `cli/charts.go:436,477,513`).

So a release whose new spec Swarm could not converge, tried, and **rolled back** reports success: `--wait` returns nil, the command exits 0, and the operator is told the deploy landed when the cluster is running the old spec. A CI pipeline gating on `swarmcli ... --wait` goes green on a failed rollout.

[...]

Issue #473, opened 2026-07-21
charts: --wait reports converged before the rollout is stable


eldara-cruncher:
`charts install --wait` / `upgrade --wait` can return success on a service that is about to crash-loop.

`Engine.waitReady` polls `allConverged` and returns as soon as the replica ratio reaches parity:

```go
// charts/release.go
func allConverged(states []ServiceState) bool {
	for _, s := range states {
		if inProgressStatus(s.Status) { return false }
		if s.Replicas == "" { continue }        // global services
		run, des, ok := strings.Cut(s.Replicas, "/")
		if !ok || run != des { return false }
	}
	return true
}
```

Two things are missing versus the algorithm the Docker CLI itself uses (`cli/command/service/progress`, which is what `docker stack deploy --detach=false` runs):

1. **No stability window.** The CLI does *not* return at parity. It holds for `UpdateConfig.Monitor` (default 5s) and resets convergence if any task fails inside that window. Swarm only counts a task failure toward `MaxFailureRatio`/`FailureAction` if it happens within `Monitor` of task creation — so a container that starts, passes the ratio check, then dies is exactly the case this misses.

2. **No down-node filter.** The CLI builds `activeNodes` from `NodeList` and excludes tasks pinned to `NodeStateDown` nodes from the denominator. Without it, a service with a task on a dead node never reaches parity and `--wait` hangs until timeout even though the rollout is as converged as it can get.

## Proposal

Port the CLI's algorithm as a pure function over `(ServiceSpec, []Task, []Node, clock)` so it is unit-testable without Docker, and have `waitReady` use it.

Note one wrinkle: the CLI filters tasks with `_up-to-date=true`, which is **undocumented** — it appears nowhere in the Engine API swagger. Safer to compute up-to-date-ness by comparing task spec against the current service spec rather than depending on an unpublished filter.

Also worth surfacing while in here: `UpdateStatus.State` of `paused` / `rollback_paused` are *terminal* failures (swarmkit never rolls back a rollback), and `UpdateStatus` is nil on a service that has never been updated — neither is currently distinguished.

Part of Eldara-Tech/swarmcli-cd#1 (Phase 0).

eldara-cruncher:
## Worse than filed — `--wait` is close to a no-op on install

While implementing this I traced the replica counts to their source, and the defect is more fundamental than the missing stability window.

`docker/service.go`:

```go
[...]

Issue #443, opened 2026-07-12
charts: one-shot (job) services break `charts status` and `--wait`


eldara-cruncher:
## Problem

A chart that renders a **one-shot service** — a task that runs to completion and is not restarted (`restart_policy: condition: none` or `on-failure`) — is permanently misreported:

- `swarmcli charts status <release>` shows it as **`0/1`** forever, which reads as "broken" to a user even though the job succeeded.
- `swarmcli charts install --wait` **never converges** and eventually fails with `timed out after <d> waiting for release <r> to converge`, even though the deploy succeeded and the release was already recorded (`record` runs before `waitReady`).

## Cause

Convergence is computed from the *desired* task state. `countTasksForNode` (`docker/service.go`) counts only tasks with `DesiredState == swarm.TaskStateRunning`, and `allConverged` (`charts/release.go`) requires `run == des`. When a one-shot task completes, Swarm sets its desired state to `Shutdown`, so the running count drops to 0 and never comes back. There is no `Complete` concept in `ServiceState`.

Swarm's native `mode: replicated-job` is not a way out: `docker stack deploy` (compose v3 schema) does not support it, and `getServiceMode` (`docker/service_helpers.go`) returns `"-"` for anything that is neither Replicated nor Global.

## Who hits it

Two charts in `swarmcli-charts` already render one-shot services, and it is the only pattern Swarm offers for an init/migration step (there is no `depends_on`):

- `ollama` — `model-init` (preloads models)
- `superset` — `init` (`superset db upgrade` + `superset init` + admin bootstrap), added in Eldara-Tech/swarmcli-charts#68

Both charts have to tell users in their README that `0/1` is the expected steady state and that `--wait` must not be used. `scripts/e2e-test.sh` in the charts repo already avoids `--wait` for a related reason (it returns while tasks are still Pending) and polls actual task state instead.

## Suggested fix

Treat a service whose tasks ended `Complete` as converged rather than as 0-of-1:

- In the convergence check, consider a replicated service satisfied when its tasks' **current** state is `Complete` (desired `Shutdown`) *and* its restart policy is `none`/`on-failure` — i.e. it is a job, not a crashed long-running service.
- In `charts status`, render such a service as `Complete` (or `1/1 (completed)`) instead of `0/1`, so a healthy stack does not look degraded.

The distinction that matters is *completed* vs *failed*: a task that exited non-zero and exhausted its restart budget should still surface as a failure.

[...]

PR #573, opened 2026-08-14
feat(charts): read-only :charts TUI browser for chart releases

### LINKED ISSUE #466: charts: read-only `:charts` TUI browser view
Carved out of #417 (charts Phase 3, optional item).

### Scope
An optional **read-only** `:charts` TUI view that reuses the `charts/` engine: search/list installed releases with their status (chart, version, revision, deploy state), mirroring `charts list` / `charts status`. Read-only — deploy/upgrade/rollback stay on the CLI.

### Current state
Only TUI *guards* exist today: `views/configs` and `views/stacks` protect chart-managed records from destructive edits (popups pointing at `charts uninstall`/`charts upgrade`). There is no browser/list view.

Lower priority than `create`/`dependency`.

eldara-cruncher:
Closes #466.

A read-only `:charts` view (alias `:chart`) for browsing installed chart releases, built on the existing `charts/` engine.

```
┌───────────────────────── Chart Releases (2) ─────────────────────────┐
│  NAME ▲          REV STATUS   HEALTH      CHART          UPDATED  UPD│
│  edge            1   deployed converged   traefik-0.1.1  …       —   │
│  hello           2   deployed progressing whoami-0.1.8   …       0.2.0│
│     REV  STATUS      CHART         UPDATED           OWNER           │
│     1    superseded  whoami-0.1.7  2026-08-11 12:00  —               │
│     2    deployed    whoami-0.1.8  2026-08-12 12:00  apply/prod-swarm│
│     SERVICE       MODE         REPLICAS   STATUS                     │
│     hello_web     replicated   1/2        preparing                  │
│  2/2                                                                 │
│  service 'hello_web': 1/2 tasks running                              │
│  Read-only · manage releases with `swarmcli charts install|upgrade…` │
└──────────────────────────────────────────────────────────────────────┘
```

`enter` expands a release in place to its revisions and services; `i`/`v` show a
revision's manifest and values, `d` diffs it against the one below it, `s` opens
the release's services. `/` filters, every column sorts, `?` explains the rest.

## What the TUI adds over `charts list`

**Live convergence health.** `charts status` already receives `[]ServiceState`
carrying `Running/Desired/Completed/Job/UpdateState/Monitor/NewestTaskAge` and
prints four display fields from it. The rules that interpret them —
`ServiceState.Convergence()` and `Rollup()` — are written, tested, carry the
scars of #443, #473, #480 and #526, and had **no presenter anywhere in the
repo**. `HEALTH` renders them, and the reason sentence `Convergence` writes for
display lands in the footer for the selected release.

`STATUS` and `HEALTH` are both present on purpose: the first is the record we
wrote when we deployed, the second is what the swarm is doing now. They agree
until they matter — a release reading `deployed` whose rollout is
`rollback_paused`. Health is rendered only for a deployed record, because
`Rollup` on no services reports "progressing", which is right for a `--wait`
deploy and wrong on a browser row for a release that failed or never installed.

**A diff between stored revisions.** Every revision record stores its rendered
manifest, and `utils/textdiff.Lines` already exists (its package doc explicitly
anticipates a TUI caller). So `d` needs no chart, no re-render and no network —
which is exactly what `charts diff upgrade` cannot say.

## Design notes

**The data seam is view-local, not a `docker.Deps` field.** `charts` imports
`docker`, so a `docker.ChartOps` returning a `charts.Release` would close an
import cycle. The interface is declared at the consumer instead, which is the
idiomatic Go placement anyway.

**One engine change:** `Engine.AllRevisions` is exported — a thin wrapper over
the existing unexported `allRevisions`. Composing `List` with a `History` per
expanded release would gunzip and YAML-decode every stored revision twice.

**The poll gate hashes the convergence phase but not its reason.** While a
service serves out its stability window the reason counts down in milliseconds,
so hashing it would report a change on every tick and save nothing.

**`UPD` never touches the network.** The repository store is built with
`RefreshNever`: this view polls, and `RepoStore`'s methods take no context, so a
fetch started here could not even be cancelled when the view closes. The indexes
are read once per view. That makes an empty cell ambiguous, so the two causes are
separated — `—` is "nothing newer", `?` is "no cached index to compare against".

**Read-only, and it says so.** A persistent footer hint; `u`/`r`/`ctrl+d` open a
dismiss-only dialog carrying the exact command for the selected release (with
the rollback target already resolved); a "Chart operations (CLI only)" help
section. Binding those keys beats leaving them dead — `ctrl+d` deletes in the
configs, secrets and volumes views, so an operator will press it here.

**`g` in `:stacks`** jumps to the release owning that stack, reusing the
ownership lookup the delete guard already makes (extracted so both share it).

## Two deviations from the plan, both found by looking at the rendered frame

The missing-index signal moved from a footer line into the `UPD` cells. With
counts, a convergence reason, the index note and the read-only hint the footer
reached four lines and squeezed the content area enough to clip a dialog on a
short terminal.

The cross-link selects and expands rather than filtering: the `/` filter belongs
to the app's search bar, and setting a query the operator never typed would
leave the list narrowed with nothing on screen saying why.

## A pre-existing bug this surfaced — not fixed here

The last commit fixes a **branching poll loop** in this view: a tick both started
a poll and re-armed the ticker, while the poll's "nothing changed" result
re-armed it too. Driving the message loop by hand, an idle view went 1, 1, 2, 3,
5, 8, 13 ticks per round — thousands of concurrent reads within a minute.

**I copied that shape from `views/volumes`.** Measured there, one tick yields two
successors in the steady state. The same shape is in `views/configs`,
`views/secrets` and `views/networks` (configs and secrets have a `polling`
atomic that prevents overlapping *reads* but not the multiplying timers).
`views/services` is the correct one — it re-arms only from the loaded handler.

Those four are untouched here: it is a pre-existing bug and deserves its own
change. Happy to open an issue.

## Testing

`go build`, all test packages, `-race`, `gofmt` and
`golangci-lint run ./... --build-tags=integration` are clean. 86.5% statement
coverage on the new package; `AllRevisions` at 100%.

Three guards were mutation-tested rather than assumed — disabling the scroll
math, duplicating a `tickCmd`, and reverting the loop fix each make the relevant
test fail.

⚠️ **`integration-tests/charts/charts_browser_test.go` is compile-checked only** —
there is no Docker daemon in my environment, so it has never been executed. It
needs `./test-setup/testenv.sh integration`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ swarmcli charts apply -f swarmcli-release.yaml --diff   # preview, never deploys
 swarmcli charts outdated                          # releases with a newer chart available
 ```
 
+The TUI has a **read-only** browser for what those commands installed: `:charts`
+lists every release with its revision, recorded status and live rollout health,
+expands one in place to its revision history and services, and diffs any two
+consecutive revisions. Everything that changes a release stays on the command
+line above; the view names the command for you.
+
 `apply` reads a release file pinning each release to a chart version, so the
 deployed state is reproducible and an automated updater (e.g. Renovate) has
 something concrete to bump. It never removes a release the file does not mention —
@@ -245,6 +251,7 @@ TEST_LOG=1 ./test-setup/testenv.sh test
 | `:secret`  | Navigate to Secret    |
 | `:network` | Navigate to Networks  |
 | `:volume`  | Navigate to Volumes   |
+| `:charts`  | Browse chart releases |
 | `l`      | View Logs               |
 | `s`      | Scale Service           |
 | `r`      | Restart Service         |
--- a/app/update.go
+++ b/app/update.go
@@ -593,6 +593,17 @@ func (m *Model) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 				return m, cmd
 			}
 		}
+		// Check if the current view has an expanded row to step out of first
+		// (charts: from a child row back to the release, then collapse it)
+		if expandableView, ok := m.currentView.(interface {
+			IsRowExpanded() bool
+		}); ok {
+			if expandableView.IsRowExpanded() {
+				// Let the view handle esc to walk back out of the expansion
+				cmd := m.currentView.Update(msg)
+				return m, cmd
+			}
+		}
 		// Check if services view is in stack services mode
 		if servicesView, ok := m.currentView.(interface {
 			IsInStackServicesView() bool
--- a/charts/README.md
+++ b/charts/README.md
@@ -39,6 +39,13 @@ swarmcli charts list
 swarmcli charts uninstall my-traefik
 ```
 
+The TUI complements this with a **read-only** browser, `:charts`: the installed
+releases with their revision, recorded status and live rollout health, expandable
+in place to the revision history and running services, with a diff between any
+two consecutive revisions. It changes nothing — install, upgrade, rollback and
+uninstall are the commands above — but it is the fastest way to see what a swarm
+is actually running, and it names the command for whatever you reach for next.
+
 A chart reference is either a configured `repo/chart` or a local path to a chart
 directory or packaged `.tgz`. **`--version` selects a version from a repository
 index, so it applies only to a `repo/chart` reference** — a local chart carries its
--- a/charts/README.md
+++ b/charts/README.md
@@ -122,7 +122,7 @@ and replaced it in 3.0 for the same reason. The `apply/` prefix keeps a manifest
 applied from the command line from colliding with a controller that happened to
 pick the same name. The controller is the other side of that: a library consumer
 passes its own id as `PlanOptions.Owner` and plans against it instead, so the
-releases it installed under `cd/<app>` read back as its own.
+releases it installed under its own `cd/…` ids read back as its own.
 
 `owner:` is optional and has **no default**. A derived one would either change
 between a laptop and a CI checkout (a path hash) or be shared by every repository
--- a/charts/release.go
+++ b/charts/release.go
@@ -711,6 +711,20 @@ func (e *Engine) List(ctx context.Context) ([]Release, error) {
 	return out, nil
 }
 
+// AllRevisions returns every stored revision, grouped by release name and
+// ascending within each group, with display statuses derived the same way
+// History derives them.
+//
+// It exists for a caller that needs the current revision of every release AND
+// the history of any of them — a browser view, where List followed by a
+// History per expanded release would decode every stored revision twice.
+// Unlike List, nothing is filtered: a release whose current revision is
+// uninstalled is present, because a caller asking for all revisions is asking
+// about the record, not about what is deployed.
+func (e *Engine) AllRevisions(ctx context.Context) (map[string][]Release, error) {
+	return e.allRevisions(ctx)
+}
+
 // Status returns the current release plus live service states for its stack.
 func (e *Engine) Status(ctx context.Context, release string) (*Release, []ServiceState, error) {
 	revs, err := e.revisions(ctx, release)
--- a/charts/release_test.go
+++ b/charts/release_test.go
@@ -556,6 +556,50 @@ func TestListReturnsCurrentRevisions(t *testing.T) {
 	require.Equal(t, "b", list[1].Name)
 }
 
+// AllRevisions is List's counterpart for a caller that needs the current
+// revision of every release AND the history of any of them from one read. The
+// two differ in exactly two ways, and both are the point: nothing is filtered,
+// and lower revisions carry their derived status.
+func TestAllRevisionsReturnsEveryRevisionUnfiltered(t *testing.T) {
+	fb := newFakeBackend()
+	e := testEngine(fb)
+	ctx := context.Background()
+
+	_, err := e.Install(ctx, "a", ReleaseChart{Name: "ca", Version: "1"}, nil, "services:\n  s:\n    image: x\n", InstallOptions{})
+	require.NoError(t, err)
+	_, err = e.Upgrade(ctx, "a", ReleaseChart{Name: "ca", Version: "2"}, nil, "services:\n  s:\n    image: y\n", InstallOptions{})
+	require.NoError(t, err)
+	_, err = e.Install(ctx, "b", ReleaseChart{Name: "cb", Version: "1"}, nil, "services:\n  s:\n    image: x\n", InstallOptions{})
+	require.NoError(t, err)
+
+	// A record another producer left in StatusUninstalled. List hides it;
+	// AllRevisions must not, because a caller asking for every revision is
+	// asking about the record, not about what is deployed.
+	gone := &Release{Name: "c", Revision: 1, Status: StatusUninstalled, Chart: ReleaseChart{Name: "cc", Version: "1"}}
+	fb.configs[releaseConfigName("c", 1)] = fakeConfig{
+		data:   mustGzipRelease(t, gone),
+		labels: map[string]string{LabelType: TypeRelease},
+	}
+
+	all, err := e.AllRevisions(ctx)
+	require.NoError(t, err)
+	require.Len(t, all, 3)
+
+	require.Len(t, all["a"], 2)
+	require.Equal(t, 1, all["a"][0].Revision, "revisions are ascending")
+	require.Equal(t, 2, all["a"][1].Revision)
+	require.Equal(t, StatusSuperseded, all["a"][0].Status, "a lower deployed revision reads as superseded")
+	require.Equal(t, StatusDeployed, all["a"][1].Status)
+
+	require.Len(t, all["b"], 1)
+	require.Len(t, all["c"], 1)
+	require.Equal(t, StatusUninstalled, all["c"][0].Status)
+
+	list, err := e.List(ctx)
+	require.NoError(t, err)
+	require.Len(t, list, 2, "List still hides the uninstalled release AllRevisions reports")
+}
+
 // Reading release history is the engine's hottest path — every List, Status,
 // Install, Upgrade, Rollback and GetRevision walks it, and a GitOps controller
 // walks it on a timer. When the Backend's listing carries the payload, that walk
--- a/commands/api/parse_test.go
+++ b/commands/api/parse_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	_ "github.com/Eldara-Tech/swarmcli/commands/command"
+	_ "github.com/Eldara-Tech/swarmcli/commands/command/charts"
 	_ "github.com/Eldara-Tech/swarmcli/commands/command/docker"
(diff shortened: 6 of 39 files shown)


mosonyi:
image


Not niceliy aligned, what is UPD?

eldara-cruncher:
Both fair — and the second one is really the first one's fault. Fixed in ba933ac.

## What UPD was

The newest version of that release's chart found in a **cached** repository
index — i.e. "you could upgrade to this". Three states:

| cell | meaning |
|---|---|
| `0.2.2` | a newer version is published |
| `—` | nothing newer, **or** the chart is in no index (a local chart) |
| `?` | no cached index at all — nothing to compare against |

The `—`/`?` split exists so "up to date" and "I don't know" can't be mistaken
for each other. It never touches the network; run `swarmcli charts repo update`
to populate it.

That is too much to infer from three letters, so it is now **`LATEST`** —
the same column name `charts outdated` already prints — and it has moved next
to `CHART`, since it is the version that chart could be at and the comparison
should be adjacent:

```
NAME          REV STATUS   HEALTH    CHART               LATEST UPDATED
openclaw      1   deployed converged openclaw-0.1.0      0.2.2  2026-08-11 12:00
swarm-cronjob 1   deployed converged swarm-cronjob-1.0.2 —      2026-08-11 12:00
```

## Why it was spread out

My bug, and your screenshot is the proof. `NAME` and `CHART` were declared
`Flex`, and `Flex` meant two different things at once:

- give up width first when the terminal is too **narrow** (and scroll with ←/→
  when truncated), and
- absorb the leftover when it is too **wide**.

Release names and chart refs are short, so on your ~200-column terminal each
flex column was handed half the slack — hence exactly the two voids you boxed,
and `UPD` stranded at the far edge. I designed and reviewed this at 100–140
columns, where the slack is small enough that it just looks like padding. At 200
it is not padding, it is a hole.

Those two elasticities are not the same property, so the shared layout now has a
separate `Grow` flag for "absorbs leftover width". `NAME` and `CHART` keep
`Flex` — a long release name still shrinks gracefully and still scrolls — and
only the trailing column grows, which puts the leftover **after** the last cell
where it reads as margin. Rows still span the full width, so the selection
highlight is unchanged.

A view that declares no `Grow` keeps absorbing on `Flex` exactly as before, so
`volumes`, `configs`, `secrets`, `networks`, `nodes` and `services` are
byte-identical. Only this view opts in.

80 columns still degrades the way it did:

```
NAME ▲       REV STATUS   HEALTH    CHART              LATEST UPDATED
openclaw     1   deployed converged openclaw-0.1.0     0.2.2  2026-08-11 12:00
swarm-cronj… 1   deployed converged swarm-cronjob-1.0… —      2026-08-11 12:00
```

## Also caught while looking

The read-only footer hint overflowed the frame at 80 columns and read as a
truncated sentence. Shortened to `Read-only · change releases with
`swarmcli charts` (? for how)`.

## Test

The regression test pins the rule, not the pixels: **every column up to the last
must sit at the same place at 100 columns and at 240.** I verified it fails if
`NAME` or `CHART` is given `Grow` again, so the void cannot come back quietly.
Plus three tests on the layout component itself, including that the no-`Grow`
fallback leaves existing views unchanged.

## One thing worth your opinion

`LATEST` currently shows a version **only when something newer exists**, and `—`
otherwise. The alternative is to always show the newest known version, which
makes the column denser but loses the at-a-glance "these three need attention".
I kept the sparse version; happy to flip it if you'd rather read it as a plain
inventory.

mosonyi:
Still not nicely aligned. Use the whole width

image

eldara-cruncher:
You're right, and the two rounds of feedback bracket the answer nicely. Fixed in
4b4cf41 — here is the reasoning, since "use the whole width" turns out to have
three quite different implementations.

## Why the obvious reading is worse

The literal fix is to spread the leftover across every column. I tried it at
your width first:

```
NAME ▲                     REV               STATUS                 HEALTH                    CHART                        LATEST               UPDATED
e2e-openclaw-no-auth       1                 deployed               progressing               openclaw-0.1.0               0.2.2                2026-08-11 12:00
keycloak                   1                 deployed               converged                 keycloak-0.1.0               0.2.2                2026-08-11 12:00
```

That uses every column and is harder to read than what you screenshotted: `1`
now sits 17 cells from its own header, and the eye has to travel the whole line
to keep one release together. It is the first complaint — the voids — multiplied
by seven instead of two.

So the two failures I shipped were the same mistake from opposite ends: treating
leftover width as something to **distribute** (round 1: holes) or to **leave**
(round 2: dead space). It is neither. It is something to **use**.

## What it does now

The column set is responsive. Each tier of surplus buys a column that has
something to say, rather than padding:

| width | columns |
|---|---|
| ~80 | `NAME REV STATUS HEALTH CHART LATEST UPDATED` |
| ~120 | … + `DETAIL` — why it is not converged, or what it is running |
| ~190 | … + `OWNER` — the `charts apply` manifest that installed it |

At your width:

```
NAME ▲               REV STATUS   HEALTH      CHART          LATEST UPDATED          OWNER                             DETAIL
e2e-openclaw-no-auth 1   deployed progressing openclaw-0.1.0 0.2.2  2026-08-11 12:00 —                                 service 'openclaw_api': 2/3 tasks running
keycloak             1   deployed converged   keycloak-0.1.0 0.2.2  2026-08-11 12:00 apply/prod-swarm:release/keycloak 1 service · 1/1 tasks running
mariadb              1   deployed converged   mariadb-0.1.1  0.2.1  2026-08-11 12:00 apply/prod-swarm:release/mariadb  1 service · 1/1 tasks running
```

and at 80, unchanged from before:

```
NAME ▲           REV STATUS   HEALTH      CHART        LATEST UPDATED
e2e-openclaw-no… 1   deployed progressing openclaw-0.… 0.2.2  2026-08-11 12:00
keycloak         1   deployed converged   keycloak-0.… 0.2.2  2026-08-11 12:00
```

Three rules make it behave across the range:

1. **A tier is bought out of genuine surplus**, measured against what the other
   columns actually need for the data on screen — never by squeezing them. A
   long release name spends the surplus, and `DETAIL` stands down rather than
   truncating the name to make room for itself.
2. **`DETAIL` is the only column allowed to give way.** Between tiers, a
   narrowing terminal eats into the explanation, not into the identity.
3. **The set is recomputed on resize and on every load**, since the widest cell
   decides as much as the terminal does.

## A side effect worth calling out

The convergence reason used to live in the footer, for the selected row only. It
now sits in `DETAIL` for **every** row when the terminal is wide enough, and
falls back to the footer when it is not. So on a wide screen you can see which
release is stuck without moving the cursor. A test pins that it is on screen
exactly once either way, so the two never both show it.

## One judgement call I'd like yours on

`OWNER` is what I picked for the second tier — it distinguishes GitOps-managed
releases from hand-installed ones, which felt like the next most useful thing
after "why is it not converged". `NAMESPACE` or a services count are the other
candidates. Easy to swap if you'd rather have something else there.

mosonyi:
image

I think this field should be swarmli-cd/...

And I should be able to move it with the left and right buttons. (Because it can be too long)

eldara-cruncher:
Both right, and the second one turned out to expose a bug in every view that
binds the arrows. Fixed in ca83339.

## The field

You are reading `cd/default/whoami:release/whoami`, which is a stamp in two
halves — `<id>:<kind>/<name>` — and both halves are a problem, for different
reasons.

**`cd/` is a product abbreviating its own name.** It is the one owner prefix an
operator cannot recognise from something they typed: `apply/prod-swarm` carries
the `owner:` *you* wrote in a release file, whereas `cd/` is what swarmcli-cd
calls itself in `application.OwnerID`. It is now spelled out on screen.

**`:release/whoami` says nothing on the row that names it.** A stamp only counts
as ownership when it names the release carrying it (that resource half is what
stops a copied stamp reading as owned — ArgoCD's `app.kubernetes.io/instance`
bug), so on `whoami`'s own row it is the NAME column again. And it is the longer
half of the longest cell in the table.

```
OWNER
swarmcli-cd/default/whoami
swarmcli-cd/default/eldara-renovate
apply/prod-swarm
—
cd/prod/gone:release/other
```

That last one is the exception, and it is deliberate: a stamp naming a
*different* release, or one that does not parse at all, is printed verbatim. It
is not evidence of owning anything on screen, so summarising it as an owner
would state something the engine itself does not believe — and it is the one
case where you came to read the exact bytes.

**Nothing stored changes.** The label on the swarm is still `cd/…`, so prune and
ownership are untouched. If you would rather the controller *wrote*
`swarmcli-cd/`, that is a swarmcli-cd change with a real migration behind it —
every already-stamped release reads as unowned until the next reconcile that
plans it re-stamps it, the same migration they ran once for `cd/<app>` →
`cd/<controller>/<app>`. Say the word and I will open the issue; it is not this
PR.

## The arrows

They work now — and the reason they did not is worth a paragraph, because the
column was one flag away.

`←/→` scroll a cell too wide for its column on the selected row, and only
columns marked `Flex` scroll. At the OWNER tier, only `DETAIL` was flex, so
OWNER truncated with nothing to press. It is flex now.

**And that flag was hiding a second bug.** `Flex` is also "gives up width first
when the terminal is too narrow" — so as a *non*-flex column, a long stamp took
its width out of `NAME` and `CHART` instead. The mutation test found it: with
the flag reverted, the assertion that fails first is not the scrolling one, it
is `the owner must start out truncated`, because the owner had helped itself to
the release's name. So the fix is the same flag both times, which is the sign it
was the right one.

## A bug this found in six other views

`ScrollRight` had no upper bound. Press `→` a few times past the end of the text
and the offset keeps climbing; once it passes the string, every flex cell on the
row renders **blank** — which reads as the data disappearing rather than as the
end of it, and takes as many presses to undo as it took to cause. That is
`volumes`, `configs`, `secrets`, `nodes`, `services` and this view.

It now stops with the last window ending exactly at the end of the text. The
clamp lives in the shared list component rather than here, since the cell width
it measures against is the component's.

Worth noting that arrow-scrolling had **zero** test coverage across the whole
repo before this.

## Also

- `←/→` is now in `?`. No view documented it anywhere, which I suspect is half
  of why you reached for it and nothing happened.
- The OWNER help entry said "the `charts apply` manifest that installed it",
  which was never the whole story and is now wrong twice over. It names both
  writers.
- One line of `charts/README.md` still described the controller's ids in the
  pre-controller-id `cd/<app>` format.

## What I did not change

The OWNER column still needs 32 columns of surplus to appear, exactly as
before. The displayed id is shorter than the stamp now, so I first lowered that
gate to 26 — and it pulled OWNER into the ~120-column tier from the last round.
Shrinking the tier boundaries is not what you asked for, so it is reverted; easy
to lower if you want the column sooner.

## Test

Three guards, mutation-tested rather than assumed: reverting `Flex`, reverting
the clamp, and reverting the stamp rendering each fail a specific assertion. I
re-ran the clamp's component test on its own, so it is not passing merely
because a sibling failed.

`go build`, all packages, `-race`, `gofmt` and
`golangci-lint run ./... --build-tags=integration` clean. 87.6% on the view.

PR #574, opened 2026-08-14
fix(views): give every polling view exactly one tick chain

eldara-cruncher:
Follow-up to #573, which found the first of these by inheriting it. Independent
of that PR — this branches off `main` and touches no file it touches.

Two commits, two defects, one rule.

## 1. The poll loop branched (`volumes`, `configs`, `secrets`, `networks`)

Each re-armed its ticker **twice per beat**:

```go
case TickMsg:
    if …ready… {
        return tea.Batch(m.checkXCmd(m.lastSnapshot), tickCmd())  // successor #1
    }
    return tickCmd()

case PollRetryMsg:      // ← what checkXCmd returns when nothing changed
    return tickCmd()    // successor #2
```

Nothing changing is the steady state of a browser left open, so this is the
common path. Driving the message loop by hand, an idle view goes
**1, 1, 2, 3, 5, 8, 13** ticks per round — within a minute it is thousands of
concurrent Docker reads, with no ceiling. Each read lists every
config/secret/network/volume on the daemon.

## 2. Then measuring the other four turned up the same rule broken again

- **Two chains from startup.** `switchToView` batches the factory's command
  *and* `OnEnter()`, and several views armed in both — or in a load handler
  reachable from either. `nodes` ran two chains for the life of the view
  (measured), as did `stacks` and `services`: twice the intended poll rate.

- **Polling stopped after a round trip.** Every view declares its own `TickMsg`
  type, and the app never broadcasts — it only calls `currentView.Update`. So a
  tick belonging to a view that is no longer current is delivered elsewhere and
  dropped: **a chain cannot survive a navigation.** Views arming only in `Init`
  — including the four from defect 1 — came back from a drill-down permanently
  stale, never polling again. `nodes` was the worst case, its `OnEnter`
  returning `nil`.

## The rule

**The tick is armed in `OnEnter`, and only a tick arms a tick.**

`OnEnter` is the one hook that runs on first entry (`switchToView`) and on every
return (`goBack`), exactly once each. `Init` and the factory no longer arm, so
entry cannot start two. Load handlers no longer arm, so a refresh cannot start
another. `PollRetryMsg` keeps its type — it is still the honest "no change"
signal — and simply stops re-arming.

Applied to all eight polling views: `volumes`, `configs`, `secrets`,
`networks`, `nodes`, `stacks`, `tasks`, `services`.

## Tests

Three regression tests per view, all of which **flatten `tea.Batch`**. That
flattening is the point, and is why this survived review in the first place: a
test asserting on the command `Update` returns cannot see past a batch, nor see
a successor that arrives one message later. (My own first attempt at this test
in #573 had exactly that flaw and passed against the broken code.)

- the loop does not multiply over 12 rounds, and still polls
- entering arms exactly one chain
- returning restarts one

Every one was mutation-tested — the fixes were reverted, in two separate views
for the entry test, and the tests fail. One pre-existing test in `views/tasks`
asserted that a load *does* arm a tick; it encoded the bug and is updated with
a comment saying so.

`PollInterval` becomes a `var` in each view, as `spinnerTickInterval` already
is in the tree: a `tea.Tick` cmd invoked synchronously blocks for the full
interval, so a test that runs one to see what it scheduled would otherwise sit
there for seconds.

## 3. Closing the gap: every tick carries its generation

The first pass left one hole open, and this branch now closes it rather than
documenting it.

"A chain cannot survive a navigation" only holds **once the tick has fired**. A
tick armed just before a drill-down can still be in flight when the operator
comes back, and it then finds the view current again and re-arms — alongside
the chain `OnEnter` just started. Nothing in the model could tell the two
apart, so the tick itself now says which entry armed it:

```go
type TickMsg struct{ Gen uint64 }

case TickMsg:
    if msg.Gen != m.pollGen {
        return nil // a leftover from an earlier entry
    }
```

`OnEnter` increments `pollGen` on every entry. A leftover is inert.

Measuring for that turned up three more arms the first pass had missed:

- **`services`' factory armed a tick of its own** (`register.go`), so entering
  the view started two chains — the exact defect §2 claims to have fixed. It
  survived because the entry test stood in for the factory with `m.Init()`
  instead of calling it. Every view's entry test now drives the **real
  factory**, which is what `switchToView` does.
- **`tasks`' factory called `OnEnter()` itself** and returned it as the load
  command, so entry loaded twice and armed twice.
- **`charts` and `contexts` had the inverse defect** — they never double-armed,
  they stopped polling. `charts` (from #573, on `main`) guards its ticker with
  a `tickScheduled` flag that only an *arriving* tick clears; a tick dropped on
  the way out of the view left the flag set forever, so `armTick` declined and
  the view never polled again after the first drill-down. `contexts` armed in
  its factory and not in `OnEnter`, which is all `goBack` runs. Both now arm
  per generation in `OnEnter`; `charts` keeps its flag, which still does its
  own job of collapsing the several paths that re-arm after a poll result.

That makes it ten polling views on one rule, not eight.

Two more regression tests per view, mutation-tested like the rest:

- a tick from an earlier entry arms no successor (fails with the guard removed:
  1 chain instead of 0)
- `charts` restarts polling on re-entry (fails without the flag reset: 0 chains
  instead of 1)

and the entry test now calls the real factory (fails with `services`' factory
arm restored: 2 chains instead of 1).

## Base

Merged `main` (`af7eb2e`) into the branch — no conflicts. `go build`, `go vet`,
`go test ./...` and `golangci-lint run ./... --build-tags=integration` are
clean.

### DIFF
--- a/views/charts/messages.go
+++ b/views/charts/messages.go
@@ -32,7 +32,7 @@ var PollInterval = 5 * time.Second
 // backstop against an unreachable daemon rather than a working budget.
 const loadTimeout = 10 * time.Second
 
-type TickMsg time.Time
+type TickMsg struct{ Gen uint64 }
 
 // PollRetryMsg signals that polling found no changes; the Update handler
 // schedules the next tick.
@@ -54,8 +54,8 @@ type ReleasesLoadedMsg struct {
 // cmd invoked synchronously blocks for the full interval.
 var spinnerTickInterval = 80 * time.Millisecond
 
-func tickCmd() tea.Cmd {
-	return tea.Tick(PollInterval, func(t time.Time) tea.Msg { return TickMsg(t) })
+func tickCmd(gen uint64) tea.Cmd {
+	return tea.Tick(PollInterval, func(time.Time) tea.Msg { return TickMsg{Gen: gen} })
 }
 
 func spinnerTickCmd() tea.Cmd {
--- a/views/charts/model.go
+++ b/views/charts/model.go
@@ -78,6 +78,9 @@ type Model struct {
 	// merely double the poll rate, they each double again on every beat.
 	tickScheduled bool
 
+	// pollGen is the generation of the live chain; see OnEnter.
+	pollGen uint64
+
 	spinner int
 
 	toastMessage string
@@ -137,7 +140,10 @@ func New(width, height int) *Model {
 func (m *Model) Name() string { return ViewName }
 
 func (m *Model) Init() tea.Cmd {
-	return tea.Batch(m.armTick(), spinnerTickCmd(), m.loadReleasesCmd())
+	// Neither the load nor the tick: OnEnter does both, and the app calls it
+	// right after the factory returns this. Doing them here as well read the
+	// swarm twice on entry and armed a second chain.
+	return spinnerTickCmd()
 }
 
 // armTick schedules the next poll, unless one is already scheduled.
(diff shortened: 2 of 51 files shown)


mosonyi:
Should I do something with it?

clangenb:
Nothing, this is a technical detail.

PR #593, opened 2026-08-17
fix(app): answer the "?" the help bar has always advertised

### LINKED ISSUE #592: The app advertises "?" globally but has no handler for it
### What happens

`app/view.go` puts `{Key: "?", Desc: "Help"}` in the global help bar for every view that does not implement `HidesGlobalHelp()`. `app/handleKey` (`app/update.go`) intercepts `esc`, `f` and `ctrl+q`/`ctrl+c` and forwards everything else to the current view — so `?` only does something if that view implements `case "?"` itself.

Nine views do. `views/logs` and `views/inspect` do not: they show the global bar, advertise `?`, and swallow the key. Nothing about the arrangement stops the next view from repeating it, which is how it happened — the container-statistics view shipped without a handler and an operator reported it (a bar promising a key that does nothing reads as a broken build, not as a missing feature).

### Suggested fix

Make `?` a real global key rather than a convention:

1. Add an opt-in interface to `views/view` — `HelpContent() []helpview.HelpCategory`.
2. Handle `?` in `app/handleKey`, respecting `IsSearching()` the way the `f` handler does (`CapturesInput()` is already checked earlier and returns before this point). Navigate to the view's own `HelpContent()` when it has one.
3. When it does not, build a generic keybinding screen from `ShortHelpItems()` plus the global keys. Every view implements `ShortHelpItems()` — it is in the `view.View` interface — so this is always available and can never be missing.
4. Drop `case "?"` from the nine views that own it today, converting them to `HelpContent()`.

That makes the defect unrepresentable rather than fixing two instances of it.

### Adjacent, same cause

- The help bar concatenates global and view entries without deduping (`views/helpbar/helpbar.go`), so the views that list `{Key: "?", Desc: "Help"}` in their own `ShortHelpItems()` render `?` twice. Step 4 above would remove the duplicates.
- Nothing in the repo binds a bare `q`, but several help screens and footers offer `esc/q` or `q/esc` for "back" — e.g. `views/charts/help.go`. Either bind it or correct the labels.

Reported against the statistics view as Eldara-Tech/swarmcli-be#326; the four affected BE views are fixed there per-view, which is the existing convention rather than a fix for this.

eldara-cruncher:
Closes #592.

### The rule

`app/view.go` advertises `f`, `?` and `ctrl+q` on every view that does not suppress the global keys. `app/handleKey` answered two of them and left `?` to each view to implement, so a view that never did swallowed it silently — a bar promising a key that does nothing reads as a broken build, not as a missing feature. `views/logs` and `views/inspect` are in that state today, and nothing stopped the next view from joining them.

`?` is now handled beside `f`, gated on `hidesGlobalKeys()` — the same condition that decides whether the bar advertises it. **The app answers exactly the keys it offers**, and a view that suppresses them gets the keystroke instead and can do what it likes with it.

### Where the content comes from

A view with more to explain than its keys implements `HelpContent() []helpview.HelpCategory`; the ten views that owned `case "?"` hand theirs over and drop the case. Every other view is described by the `ShortHelpItems()` it already publishes to the bar — that is part of the `View` interface, so it is never missing, and a view written tomorrow answers `?` on the day it is written without anyone remembering to wire it.

`globalHelpEntries()` is extracted from `View()` so the bar and the screen are built from one list; a test asserts both surfaces carry every entry in it.

### Two things it does not change

`?` while a view's internal search is open is still a character, guarded exactly as `f` is two lines below. I read every `CapturesInput` and `IsSearching` implementation in `views/`: every text-entry mode in the repo — create dialogs, label input, import input, node select, scale dialog, logs' and inspect's search — is behind one of them, and `CapturesInput` already returns before this point.

`?` in the used-by sub-views of `configs`, `secrets` and `networks` stays inert. Those report `IsSearching() == true` while the sub-view is open, so the key is forwarded — but `handleUsedByViewKey` intercepts before the normal switch, so it did nothing there before this change either. Worth fixing one day by not conflating "in a sub-view" with "typing"; not worth widening this change for.

### Also from the issue

The nine views that listed `{Key: "?", Desc: "Help"}` in their own `ShortHelpItems` rendered it twice beside the app's entry — `helpbar.View` concatenates global and view entries without deduping. Removed, and the authoring comment in `views/help/model.go` now says not to add it.

Six help screens offered `<esc/q>` for "back". Nothing in this repo binds a bare `q` (`ctrl+q` quits), so they now say `<esc>`.

`docs/features.md`'s container-statistics key table gains the `?` row it was missing.

### Tests

`app/help_test.go` covers the four behaviours: a view that says nothing gets a screen built from its keys; a view with `HelpContent` gets its own; a searching view receives the keystroke; a view hiding the global keys receives it too. Plus a guard that the bar and the screen agree on the global list, and unit tests for `FromKeys`. Each was checked by reverting the code under it — dropping the searching guard, the `hidesGlobalKeys` gate, and the `?` entry each fail exactly one test and no others.

The ten views' `TestKey_Help` tests asserted a mechanism this removes; they now assert what the view still owns, which is that it supplies help content. `newTestAppModel` gained the two dialogs and the header it was missing, so `View()` can be called from a test at all.

### Downstream

Business Edition has three views that own `case "?"` as of this morning. They are shadowed by this and fall through to the generic screen until they are converted to `HelpContent()`; a follow-up PR does that. Released BE pins a CE tag, so nothing shipped is affected.


mosonyi:
image

Can't read the whole sentence

eldara-cruncher:
Fixed in #594 — and it was worse than the screenshot showed. The same arithmetic panics on a 102-column terminal with that exact sheet, renders 324-character lines below ~110 columns, and can split a rune into invalid UTF-8. The layout now wraps on display width and packs categories into as many columns as the terminal has room for, and the sheet scrolls when it outgrows its frame. Thanks for catching it.

PR #594, opened 2026-08-17
fix(views): wrap the help sheet, and say ✓ when a release is up to date

eldara-cruncher:
Follow-up to #593 — https://github.com/Eldara-Tech/swarmcli/pull/593#issuecomment-5315138544, "Can't read the whole sentence".

### What the screenshot was of

`buildCategorizedContent` took **one column per category**, each `terminalWidth / numCategories` wide, with a fixed 15-column key field, and cut the description to whatever was left. Six categories on a 200-column terminal leaves about seventeen columns of prose, hence `The recorded state of the release: wh...`.

The column count was a property of the content, not of the terminal: every category an author added took space away from all the others, and the only defence left was the cut.

### Three more things fell out of the same arithmetic

I probed every `GetXHelpContent()` in the repo from 60 to 300 columns.

- **It panics.** The charts sheet at width 102: `slice bounds out of range [:-2]`. The remaining space lands on one or two columns and the code slices `Description[:descWidth-3]`. Shipped content, ordinary terminal size.
- **It garbles narrow terminals.** At width 80 the same sheet renders content lines **324 characters wide**, because when the remaining space is ≤ 0 the cut is skipped and nothing constrains the text at all. The header row reads `THE LATEST COLUMNSORTING`.
- **It can emit invalid UTF-8.** The cut counts bytes, so it splits multi-byte runes; 8 of 12 offsets I tried on an em dash produced a `\xe2\x80` fragment. That sheet contains `—`, `←/→` and `↑/↓`.

Nine of the ten sheets truncated at 120 columns or less; `charts` truncated at every width up to 300 — its longest description would need 840 columns to fit the old layout.

### The fix

Categories are rendered as blocks and wrapped on **display width** (`lipgloss.Style.Width`, which never splits a rune), with the key column sized per category rather than a global 15 — and capped at half the block, so a category documenting values rather than keystrokes cannot leave its descriptions a sliver. Under that cap an item stacks: key on its own line, description indented below, the same fallback the per-command help already uses.

Columns are packed to fit the terminal: `width / 44`, capped at the number of categories. The split is the one that **minimises the tallest column**, found by binary search over the ceiling. Filling each column to an average and moving on does not work — whatever is left when the last column starts has to go in it, which is how the first attempt ended with three categories stacked beside three empty ones. If the balance settles on fewer columns than the width allows, the layout runs again in that many, so the columns that remain get the whole width.

Wrapping makes a screen taller than its frame possible, so the content now goes through `m.Viewable` — the viewport that was already on the model and already receiving ↑/↓/pgup from `Update`, but was never given this content. The footer offers the scroll keys when there is something to scroll to. **Before this, a sheet that outgrew its frame lost the rest with no indication.**

### Tests

`TestLayout_KeepsEveryWord` is the assertion the screenshot was of: for every sheet in the repo, at a width that forces one column, each description survives with its wrapping whitespace removed. `TestLayout_StaysInsideTheTerminal` renders all ten across eight widths — including 102 and 108 by name, where the old code panicked — asserting valid UTF-8 and no line wider than the terminal. Plus the wrap alignment, the phrase-shaped key, scrolling to the tail, and `TestPackColumns_IsTheBestSplitAvailable`, which checks the binary search against an exhaustive search of every contiguous split.

Each guard was checked by reverting the code under it. Worth noting one that did *not* bite at first: with the column padded to its width at the end, text that is never wrapped gets folded there anyway and stays inside the terminal — so the width test alone could not tell wrapping from no wrapping. `TestLayout_WrapsUnderTheDescription` closes that by asserting continuation lines start at the description column rather than under the key.

### Downstream

Business Edition's three sheets render through this. I ran its suite against this branch: green, and the container-statistics sheet — whose keys include a 25-rune phrase — lays out in three balanced columns with every sentence whole.

---

## Second commit: the LATEST column says ✓ when there is nothing to install

From the review — https://github.com/Eldara-Tech/swarmcli/pull/594#issuecomment-5315486030.

`—` stood for two different facts: *nothing newer is published*, and *this chart is in no cached index, so there is nothing to compare against*. An operator acts on those differently — the first is done, the second means a local chart or an uncached repo.

**The distinction was discarded a layer down.** `charts.Outdated` skips a release already at the newest version and one whose chart is in no index alike (`bestVer == "" || compareVersions(...) <= 0 → continue`), so the view could not tell them apart from what it was handed. `charts.Available` now reports what the indexes know about every release whose chart is in one — `Latest` plus a `Newer` flag — and `Outdated` is that filtered to the upgrades. Same entries, same signature, so `swarmcli charts outdated` is untouched; a test pins the two together so the filter cannot drift from the source.

The column now reads:

| cell | meaning |
|---|---|
| `0.2.0` | a newer version is published |
| `✓` | in a cached index, nothing newer |
| `—` | in no cached index — a local chart, nothing to compare against |
| `?` | no cached index at all, run `swarmcli charts repo update` |

A release *ahead* of every index reads `✓` too: there is nothing newer to install, and offering a downgrade in a column called LATEST would be worse than saying nothing.

The charts help screen documents all four — which is what ties this to the first commit, since that fourth row is only readable because the sheet now wraps.

While updating the view's test double I made it answer through the real `charts.Available` over a constructed index rather than re-implementing the rules by hand; the hand-rolled version compared versions with string equality, which is exactly how a fake comes to disagree with production.

### DIFF
--- a/charts/outdated.go
+++ b/charts/outdated.go
@@ -5,31 +5,36 @@ package charts
 
 import "sort"
 
-// OutdatedEntry is one installed release with a newer chart version available.
-type OutdatedEntry struct {
-	Release   string
-	Chart     string
-	Repo      string
-	Installed string
-	Latest    string
+// Availability is what the cached repository indexes know about one installed
+// release's chart. It exists because "nothing newer" and "nothing to compare
+// against" are different answers that an absent entry ran together: a local
+// chart is in no index, and a release at the newest published version is in one
+// — the second is up to date, the first is merely unknowable.
+type Availability struct {
+	// Repo is the repository that supplied Latest.
+	Repo string
+	// Latest is the newest version of this chart in any index, whether or not
+	// it is newer than what is installed.
+	Latest string
+	// Newer reports that Latest is an upgrade from the installed version. A
+	// release ahead of every index — installed from a local chart, say — is
+	// current rather than outdated, so this is false there too.
+	Newer bool
 }
 
-// Outdated joins installed releases against the newest version of their chart in
-// any configured repository index. Releases already at the newest version, and
-// those whose chart appears in no index (a local chart), are omitted.
+// Available joins installed releases against the cached repository indexes,
+// keyed by release name. A release whose chart appears in no index is absent
+// from the result, which is the one fact Outdated cannot report.
 //
 // A Release does not record which repository it came from, so a chart present in
 // two repositories resolves to the highest version across them, reporting the
 // repository that supplied it. That ambiguity is documented rather than designed
 // away: recording the source repository is a change to persisted release state,
 // which is not worth making for a case most users never hit.
-func Outdated(rels []Release, indexes map[string]*Index) []OutdatedEntry {
-	var out []OutdatedEntry
+func Available(rels []Release, indexes map[string]*Index) map[string]Availability {
+	out := make(map[string]Availability, len(rels))
 	for _, rel := range rels {
-		var (
-			bestVer  string
-			bestRepo string
-		)
+		var bestVer, bestRepo string
 		for repo, idx := range indexes {
 			versions := idx.Entries[rel.Chart.Name]
 			if len(versions) == 0 {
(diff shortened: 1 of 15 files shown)


mosonyi:
If the version is up2date do not show a - but a Check mark: ✓
Local or without version can be -

eldara-cruncher:
Folded in as `856679f`. `✓` when the chart is in a cached index and nothing newer is published, `—` only when there is nothing to compare against (a local chart, or a repo you have not cached), `?` when no index is cached at all.

It needed a change a layer down: `charts.Outdated` skipped "already newest" and "in no index" alike, so the view was never given the difference. `charts.Available` now reports what the indexes know about every release whose chart is in one, and `Outdated` is that filtered to the upgrades — `swarmcli charts outdated` is unchanged.

One judgement call worth flagging: a release *ahead* of every index also reads `✓`, since there is nothing newer to install and offering a downgrade in a column called LATEST would read worse than saying nothing.

```
NAME      REV  STATUS    CHART              LATEST
current   1    deployed  c2-3.0.0           ✓
local     1    deployed  unpublished-0.1.0  —
stale     1    deployed  c-1.0.0            2.0.0
```

mosonyi:
LGTM

PR #597, opened 2026-08-19
docs(license): managed licensing, and an honest privacy section

eldara-cruncher:
Public documentation for managed licensing. The engineering half is `Eldara-Tech/swarmcli-be#279`; this is the customer-facing half, and it lands here because `docs/license.md` in this repository is where that page actually lives — the copy in `swarmcli-be-releases` is a leftover (see the note at the end).

## The part that needs a human decision, not a merge

**§ Privacy changes a published commitment.** It said:

> No network call is made for license validation, and none is possible — an air-gapped swarm activates exactly like a connected one.

Validation is still exactly that, and the section now says so more precisely than before: swarmcli checks the signature itself, on every kind of license, so no outage of ours can disable anybody's features. **What changes is renewal.** A managed license is activated per swarm with a lease that expires, so obtaining the next lease means asking us.

The rewrite states exactly what that request carries — **license id, cluster id, swarmcli version, timestamp** — and what it does not, which is everything read from a customer's Docker daemon. It also states the two things people actually ask about:

- the request is made by **the operator's swarmcli client**, not by anything deployed into their cluster — the agent and proxy have no outbound path to us;
- it is **not required**: a lease can be delivered as a file (`:license lease install <file>`), which is the supported path for air-gapped and policy-restricted clusters rather than a workaround.

And licenses that are unbound or bound at issuance still make no such request, ever.

Please read that section as a commitment rather than as prose. I have tried to make it narrower and more checkable than what it replaces, but "we now sometimes make a network call" is the kind of sentence a customer quotes back.

## Everything else

- **Model** — the phones-home claim reworded to be true of both kinds, with a pointer to Privacy rather than a footnote.
- **Activation** — managed licenses documented as the two-step thing they are: install the key, then activate each swarm with a lease. Includes the 30-day renewal + 30-day grace default and the advice to ask for a long lease *before* a first outage rather than after.
- **Per-swarm binding** — three modes named (unbound, bound at issuance, managed), and a section on moving a managed license: an account action, rate-limited and recorded, with **no client command and why**. A key is exportable on purpose, so if holding one were enough to move it, a copied key would be enough to take a swarm's license away from it.
- **My cluster was rebuilt** — materially better answer for managed licenses: the key you hold is still your key, ask for the license to be moved and install the lease. The bound-at-issuance answer is unchanged.
- **Lifecycle states** — the managed states get their own diagram, and the state table gains four rows. `Not activated` and `Activation expired` are called out as different problems with different fixes, only one of which involves us.
- **License id** — what it is, why support asks for it, and that it is not a credential.
- **`:license` subcommands** — `id`, `lease install`, `lease show`, plus a line saying there is no subcommand that moves a license.
- **Renewing** — the key and the activation are separated, since they are now different operations.

Managed licenses are flagged as *being rolled out*: keys issued today are unbound or bound at issuance and nothing about them changes.

## Deliberately not here

No key hierarchy, no `kid`, no Config names for the lease, no clock handling, no deny-reason vocabulary, no threat analysis. Those live in `swarmcli-be/docs/managed-licensing.md`, and the split is the usual one — mechanism and operator procedure here, internals there.

## Two things found while writing

**`swarmcli-be-releases/docs/license.md` still publishes a full copy of this page**, including the old privacy promise, at 282 lines. There is an unpushed local branch (`docs/point-at-the-one-home`) that replaces it with a pointer here, which is the right fix and is not mine to push. Until it lands, the old promise stays published in a second place — worth resolving in the same pass as this PR.

**The staleness `swarmcli-be#279` predicted is already gone.** That issue says § Activation still documents env-var and file auto-loading in priority order; it does not — it correctly says neither activates on its own. No change needed, recorded here so nobody goes looking.

### DIFF
--- a/docs/license.md
+++ b/docs/license.md
@@ -12,9 +12,15 @@ goes through.
 ## Model
 
 A license key is a single opaque string, cryptographically signed by Eldara
-Tech and verified by the `swarmcli` binary offline — nothing phones home, and a
-key works on an air-gapped swarm. Keys are versioned, so a key issued today
-keeps working when the format gains a field.
+Tech and verified by the `swarmcli` binary offline. Verification never involves
+us: swarmcli checks the signature itself, so a key works on an air-gapped
+swarm and no outage of ours can disable your features. Keys are versioned, so
+a key issued today keeps working when the format gains a field.
+
+A [managed license](#managed-licenses-activation-is-a-second-step) adds a
+renewal step that *can* use the network — never verification, only renewal, and
+there is an offline path for it. [Privacy](#privacy) says exactly what that
+sends and when.
 
 What a key records about you:
 
@@ -23,8 +29,10 @@ What a key records about you:
   feature set today; `trial` is meant to be paired with a short expiry.
 - **When it expires**, if ever. A key may be issued without an expiry.
 - **Node and user limits**, if any. See [Limits](#limits).
-- **Which swarm it is bound to**, if any. See
+- **Which swarm it is bound to**, if any, and **how** — see
   [Per-swarm binding](#per-swarm-binding).
+- **A license id** — a short `lic_…` string naming this individual license.
+  See [License id](#license-id).
 
 What a key does *not* record is a list of features. Entitlements are derived
 from the tier, so adding a capability to a tier benefits every outstanding key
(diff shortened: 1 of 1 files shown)


eldara-cruncher:
**Correction pushed (`cabcef9`) — please read § Privacy as it stands now, not as first written.**

The first draft said the license key "never leaves your machine … during activation, renewal, or anything else". That is true of every license type except the one the section exists to describe. A managed renewal presents the token as its **bearer credential** (`swarmcli-be/docs/managed-licensing.md` §3: `Authorization: Bearer <full licence token>`), so on that one path the key *is* transmitted.

The claim is now split rather than softened:

- unbound and bound-at-issuance keys never leave the machine — there is nothing to renew, so nothing is sent;
- a managed renewal sends the key over TLS as proof the request is yours, and the section says so plainly, along with why it discloses nothing new: we issued and signed that key.

The list of what the request carries now includes the key itself, alongside the license id, the cluster id, the swarmcli version, and the ordinary time-and-origin of an HTTPS call.

I am flagging this rather than quietly amending because it is exactly the failure mode this PR is meant to avoid. A privacy section that has to be walked back after a customer quotes it is worse than one that reads less cleanly today, and I wrote the over-claim while trying to make the section reassuring — which is the direction this kind of text always drifts.

If the transmitted-key property is unwanted, that is a **contract** decision rather than a docs one: §3's credential model would have to change to something that proves possession without sending the token (a signature over a challenge, say), and both this page and the issuer would follow. Worth deciding before the issuer implements the endpoint rather than after.

PR #603, opened 2026-08-20
demo: a re-recordable fixture and VHS tapes for the product demo

eldara-cruncher:
Draft: the machinery for a new product demo. Nothing is recorded yet — this is the
fixture and the tapes, so the recording session has something to check out and so the
next one is a re-run rather than a re-invention.

Design and decisions: swarmcli-website#85.

## Why

The demo in this README and on the landing page is the same file (`md5 3cc74413…`), one
32.7 s Terminalizer capture nobody can reproduce. It shows **v1.1.0**, a **one-node swarm
called `docker-desktop`**, swarmcli's own infrastructure as the workload (`stack_agent`,
`stack_proxy`), a logs pane with five identical `agent on :8080` lines, and it ends in
`nano` on a real `haproxy.cfg` — after showing `mosonyi/swarmcli-agent:local` and a WSL
bind-mount path on screen. The header carries the version, so every release ages it.

## What is here

| File | Role |
|---|---|
| `stack-shop.yml` | the workload on screen — replicated + global services, deterministic app-like log lines, **no `deploy.resources` limits** (DinD refuses to start one) |
| `swarmcli-release.yaml` | three chart releases; `traefik` and `whoami` pinned behind their published versions so the charts view's LATEST column shows real upgrades beside a `✓` |
| `docker-compose.demo.yml` | overlay on `test-setup/` that names the three DinD nodes, which otherwise appear as container IDs |
| `stack-stats.yml` | the container-statistics fixture for a **real** daemon: a 128 MB limit so the pane reads `MEM / 128MB`, plus a load loop that moves all four panes |
| `setup.sh {dind\|stats}` | puts either cluster into that state |
| `*.tape` | the recordings; `hero.tape` and `readme-gif.tape` both `Source demo/beats-hero.tape`, so the video and the GIF cannot drift |

## Two environments, and why

The DinD test swarm runs on a threaded cgroup v2 hierarchy: `memory` and `io` are domain
controllers and cannot be enabled there, so the stats view's MEM and BLK panes read
*"not reported by this host"* for every container, and a service declaring a memory limit
does not start at all. So the hero and charts clips come from the 3-node DinD swarm — a
real multi-node cluster on screen, at last — and the stats clip from a real single-node
daemon. `setup.sh stats` refuses to run where a memory limit cannot be enforced, which is
that condition checked directly rather than trusted to a comment.

## Verification

Off a Docker host, so: all four YAML files parse; `swarmcli-release.yaml` parses through
this repo's own strict `charts.ParseReleaseFile` (3 releases, wave 1 on `whoami`); the
embedded shell scripts pass `sh -n` and emit the intended log lines when run; all five
tapes pass `vhs validate`, with a deliberately broken tape as a control so the validator
is known to have been awake; `scripts/check-spdx.sh` exits 0.

**Not verified:** the sleep timings, and the `Down 3` that selects `shop` on the stacks
list — that depends on the view's sort order and is commented as needing a check on the
first dry run. That is what keeps this a draft.

## CI

`demo/**` matches neither `ci.yml`'s `code`/`docker` filters nor `integration-tests.yml`'s
`integration` filter, so this runs no build and no DinD suite. The SPDX job does fire —
its filter includes `**.sh` — and passes. When the recording lands, the same PR picks up
`assets/swarmcli.gif` and the README alt text, and `README.md` *is* in the `code` filter,
so full CI runs then, which is what the charts-help-block test is there for.

https://claude.ai/code/session_014HhwELQEsHQ9gvfJzv8mrj


mosonyi:
Please move this PR to the swarmcli be.

eldara-cruncher:
Moved to swarmcli-be#380, closing this one.

GitHub has no transfer for a pull request — `transferIssue` takes an Issue node and this
one's is a `PullRequest` — and the two repositories share no history, so the tree was
re-landed there rather than pushed across.

Four things changed in the move, and nothing else:

- `setup.sh` reaches `test-setup/docker-compose.yml` through the sibling checkout
  (`../swarmcli`), which is how swarmcli-be's own `testenv.sh` already resolves it.
  Compose names the project after that file's directory, so it is still the same swarm.
- The README GIF is written to `demo/out/` instead of straight to `assets/swarmcli.gif`.
  That asset is this repository's, and swarmcli-be is private, so recordings are copied
  out rather than committed where they are made.
- The Apache-2.0 SPDX headers are dropped: swarmcli-be carries none and has no SPDX job.
- `demo/README.md` says all of the above.

Worth knowing for the recording session: the assets still land **here** and on the
website. `assets/swarmcli.gif` and the README alt text are a pull request against this
repository, and the hero video, poster frame and feature clips go to swarmcli-website.
Only the fixture moved.

The `demo/recording-fixture` branch here is left in place, unmerged — say the word and
I will delete it.

PR #610, opened 2026-08-24
feat(logs): a separator you can put in, and new lines that announce themselves

### LINKED ISSUE #608: Log viewer enhancements
When I watch the log with tail, sometimes I need to wait until the log is coming.
I want to add some enter (empty lines), to visually separate the old and new ones.

Like in bash ```tail -f log.txt```

The next request is that the new logs are highlighted/fading just for a second and they go back to normal.
Similar to https://github.com/Nangal/TailBlazer but use the original colors maybe in bold.

eldara-cruncher:
Closes #608.

Two things a reader watching a `tail` asked for, and could not have here.

### A separator you can put in

Under `tail -f` you hit enter to push what you have read up the screen, so
whatever arrives next is unmistakably new. There is no scrollback to push in a
viewport, so `enter` — until now an explicit no-op in the logs view's normal
mode — puts the break into the buffer instead:

```
web.a1b2c3d4e5f6@node-1 | backlog request served 2

─────────────────────────────── 13:05:48 ───────────────────────────────
web.a1b2c3d4e5f6@node-1 | LIVE request served 9        (banded, for 1.2s)
```

It is stored as a sentinel rather than as rendered text, so a resize re-cuts the
rule instead of leaving one from the old width behind. It is exempt from the
node, `/` and hide-stopped filters — a filter that swallowed it would answer the
keypress with nothing at all — and from the horizontal scroll, since a separator
that slid out of frame leaves nothing to have marked. It is not a log line, so
it stays out of the count in the title and out of the search matches: a query of
`0` must not land on a timestamp the reader did not write.

### A highlight that expires

A line that has just arrived is drawn over background `24` — the colour a
picked-out row carries elsewhere in the app — for 1.2s, and then goes back to
exactly what it was, keeping the colours it arrived with. Only the background is
set: the foreground stays whatever the line arrived with, which is what the issue
asks for, and a dark background keeps a container's own red, dim or bright text
readable on it.

*(This was bold until review: @mosonyi found it not recognizable enough, which it
is not — half the lines a service logs are bold somewhere already.)*

The band is re-asserted past everything the line does that would clear it
(`ui.BackgroundANSI`): every reset — our own node prefix ends in one, and a plain
`\x1b[48;5;24m` + line would tint the text only as far as that — and `\x1b[49m`,
which restores the default background without resetting anything else. The
arguments of an extended colour are stepped over rather than read: the `0` in
`38;5;0` is a colour index, not a reset, and a line that set its own background
has to keep it across one.

The band covers the whole row, padded to the viewport width, so it reads as a
band and not as a differently-tinted sentence. That has to be the last thing
done to a row: padding measured before the horizontal scroll comes up short by
the offset, and a band painted before the wrap stops at the first break, leaving
the rest of a wrapped line looking old. `renderRows` therefore only records which
rows are fresh. Wrapping moved from one pass over the joined buffer to one per
row — after a single wrap of everything, which output row came from which line is
no longer knowable — and a row that already fits is not wrapped at all, which is
nearly all of them. The search highlight ends in a reset the band is re-asserted
past, and the ordering that used to be a rule inside `renderRows` is now
structural.

A stream opens with the backlog Docker replays for the tail request. That is
history, not news, and highlighting it would flash the whole screen at open. The
highlight therefore arms only once the replay is over, on either of two signals,
each covering the other's blind spot: a gap between arrivals, which a quiet
service gives immediately, or more lines than the backlog can hold, which is the
only signal a service logging faster than the gap will ever give.

The beat that expires a highlight is armed once and re-arms only while something
is still fresh — nothing ticks at rest, and a burst of lines cannot give one
beat a successor per line.

### Groundwork

`lines`, `lineNodes` and `lineTasks` were appended and trimmed in lock-step at
four sites, and this adds two more slices to them (kind, arrival time). Both
operations now go through `appendLine` / `trimToMaxLines`, so five parallel
slices are only ever grown and cut in one place, and the viewport bookkeeping
that follows a trim is `syncViewport` rather than a copy per caller.

### Not done, deliberately

No automatic separator after a quiet stream, no toggle key, no persisted
setting, and no true colour-interpolated fade — interpolating a foreground
colour means overriding the colours the issue asks to preserve, and it costs a
full rebuild of a 10 000-line buffer every animation frame.

### Verification

`go test ./...`, `golangci-lint run ./... --build-tags=integration`, `go vet`
and `go test -race -count=2` are all clean. Every new behaviour was
mutation-checked: nineteen mutations (drop the filter exemption, drop the band,
drop the blank row, drop the scroll exemption, count marks as lines, treat marks
as searchable, stamp the backlog, arm per line, never stop the beat, look past
marks in `anyFresh`, ignore the trim in the offset, drop the re-assertion, drop
the padding, paint before the scroll, wrap the joined buffer instead of row by
row, one `false` per mark row instead of two, drop the `49` case, read an
extended colour's arguments instead of stepping over them, and step over a
fixed number of them) each failed the test written for it, and none survived.

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/charmbracelet/lipgloss"
 	"github.com/charmbracelet/x/ansi"
+	"github.com/muesli/termenv"
 )
 
 // Styles (you can override these per-view if desired)
@@ -433,24 +434,23 @@ func TruncateANSIAfter(s string, skipWidth int) string {
 	return ansi.TruncateLeft(s, skipWidth, "")
 }
 
-// SGR escapes used to embolden a line that already carries its own colours.
-const (
-	sgrBold  = "\x1b[1m"
-	sgrReset = "\x1b[0m"
-)
+// sgrReset ends the attribute this file asserts over a line.
+const sgrReset = "\x1b[0m"
 
-// BoldANSI renders s bold without discarding the colours it already carries.
-// A log line arrives with its own escapes — our own node prefix ends in a reset
-// — and a reset clears the bold along with everything else, so a plain
-// sgrBold+s would embolden the text only as far as the first one. Bold is
-// therefore re-asserted after every reset the line contains.
-func BoldANSI(s string) string {
+// BackgroundANSI draws s over background colour bg without discarding the
+// colours it already carries. A log line arrives with its own escapes — our own
+// node prefix ends in a reset — and a reset clears the background along with
+// everything else, so a plain sequence+s would tint the text only as far as the
+// first one. The background is therefore re-asserted after everything that
+// turns one off.
+func BackgroundANSI(s string, bg termenv.Color) string {
 	if s == "" {
 		return ""
 	}
+	sgrBg := termenv.CSI + bg.Sequence(true) + "m"
 	var b strings.Builder
-	b.Grow(len(s) + 2*len(sgrBold) + len(sgrReset))
-	b.WriteString(sgrBold)
+	b.Grow(len(s) + 2*len(sgrBg) + len(sgrReset))
+	b.WriteString(sgrBg)
 	for i := 0; i < len(s); {
 		seq, ok := csiAt(s, i)
 		if !ok {
@@ -460,9 +460,10 @@ func BoldANSI(s string) string {
 		}
 		b.WriteString(seq)
 		i += len(seq)
-		// seq is "\x1b[" + params + final; re-assert only after an SGR reset.
-		if seq[len(seq)-1] == 'm' && sgrResets(seq[2:len(seq)-1]) {
-			b.WriteString(sgrBold)
+		// seq is "\x1b[" + params + final; re-assert only where the line has
+		// turned the background off.
+		if seq[len(seq)-1] == 'm' && sgrClearsBackground(seq[2:len(seq)-1]) {
+			b.WriteString(sgrBg)
 		}
 	}
 	b.WriteString(sgrReset)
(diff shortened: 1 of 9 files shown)


eldara-cruncher:
Second commit adds a fix for a defect found in the code this PR already
touches, on the two paths that append a stream notice (`StreamErrMsg`,
`StreamDoneMsg`).

Both appended their line as an ordinary log line with an empty node and task,
and hand-rolled the viewport update. Probed on the pre-fix code:

```
follower after StreamDoneMsg:  atBottom=false  last visible row = "line 19"
with node filter "node-1":     content = "real log"        <- error not shown at all
with "/nginx" active:          content = ""                <- notice not shown at all
MaxLines=2 after two notices:  buffer holds 4 lines
```

The node-filter case is the one that matters: the filters key off a node and a
task a notice does not have, so a reader filtered to one node watched the
stream die and was told nothing.

Notices are now a `lineKind` of their own — exempt from the filters exactly as
the separator is, and out of the count in the title, since the view speaking is
not the service logging. They stay matchable by `ctrl+f`, which seeks rather
than hides. Both paths go through `syncViewport(trimToMaxLines())`, so all
three append sites are identical and `lineVisible` states the rule once.

Six mutations were run against the new tests. Two survived on the first pass —
the follower and trim tests exercised `StreamDoneMsg` only, and `StreamErrMsg`
is a separate case in `Update` — so both became table tests over the two
messages. All six are caught now, each by its own subtest.

mosonyi:
Almost good, but the highlight is not really recognizable. Maybe highlight the background instead of make the text bold

mosonyi (changes requested):
See comment above

eldara-cruncher:
Agreed, and done in 1b25e6b — bold was the wrong instrument. Half the lines a
service logs are bold somewhere already, so the weight difference had nothing to
land against.

A new line is now drawn over background `24` (the colour a picked-out row carries
elsewhere in the app) for the same 1.2s, and the band covers the **whole row**,
padded to the viewport width — a background that stopped where the text stopped
would have been about as recognizable as the bold was.

Only the background is set; the line keeps the foreground colours it arrived
with, which is what #608 asks for, and `24` is dark enough that a container's own
red, dim or bright text stays readable on it.

Two things fell out of it that are worth knowing about:

- The band has to be the **last** thing applied to a row. Padding measured before
  the horizontal scroll comes up short by exactly the offset, and a band painted
  before the wrap stops at the first break, leaving the rest of a wrapped line
  looking old. So wrapping now happens per row rather than over the joined
  buffer — after one wrap of everything, which output row came from which line is
  no longer knowable.
- Re-asserting past a reset is no longer enough. `\x1b[49m` restores the default
  background without resetting anything else — invisible to a bold, fatal to a
  band — and the arguments of an extended colour now have to be stepped over
  rather than read: the `0` in `38;5;0` is a colour index, and a line that set its
  own background must keep it across one.

Both are mutation-checked, along with the padding and the paint order. `go test
./...`, `-race -count=2`, `go vet` and `golangci-lint` are clean.

If `24` reads as too quiet on your terminal it is a one-line change
(`freshBackground` in `views/logs/view.go`) — say the word and I will swap it.

mosonyi:
LGTM, please merge it!

clangenb (approved):

Issue #563, opened 2026-08-12
ValidateContext validates the context being left, not the one being switched to


eldara-cruncher:
`docker.ValidateContext` switches to the target context and then checks the connection through `GetClient()`. That returns a **process-wide cached client**, and nothing drops it first — so the ping and the swarm-state check are performed against the context the user is *leaving*.

Inside a running TUI a client is always already cached, so `ValidateContext` returns `nil` regardless of whether the target is reachable. `:contexts` therefore never actually validates a switch. It looks like it works only because the app-level snapshot load fails afterwards and reverts (`app/update.go`), which reports a vaguer reason and after a visible detour through the loading view.

`docker/client.go` documents the requirement in as many words — "Call `ResetClient` to force a fresh client (e.g. after a context switch)" — and `ValidateContext` is the one caller that switches and does not.

**Reproduce**

1. `docker context create dead --docker host=tcp://127.0.0.1:1`
2. Run `swarmcli` against any working context, so a client is cached.
3. `:contexts` → select `dead`.

Expected: the switch is refused with a connection error. Actual: it is accepted; the app then fails loading the snapshot and reverts.

**Second defect in the same function**

`DOCKER_CONTEXT` is resolved ahead of `~/.docker/config.json` (`GetContextFromEnv`), so when it is set to some *other* context, `docker context use` writes a file this process ignores and the switch cannot take effect at all — yet `ValidateContext` reports success. Setting it to the same context being validated is not a conflict and should keep working (the locked-swarm integration test relies on exactly that).

Both are the same user-visible lie: "switched" when nothing moved, or "reachable" when nothing was checked.

**Note on the fix**

Correcting this changes `:contexts` behaviour: switches that today succeed and are then quietly reverted by the snapshot path will start failing up front in the error dialog, with the real reason. That is the point, but it is worth calling out in the release notes.

PR #612, opened 2026-08-26
fix(context): one Docker context per session, and a prompt when the shell leaves it

### LINKED ISSUE #611: Context switch command line not discovered
When swarmcli is opened and I switch the docker context in the backend with
```docker context use another-context```

Swarmcli doesn't recognise it and then the I got failures when for example want to see logs.

Suggestions
1. when I use context switch in swarmcli, it should use it as a session variable and if in the backend I change context it should not have an influence on it.
2. Realize the context switch and show a popup the the context has changed from x to y, proceed or roll back to x?

### PR DESCRIPTION
**Pull request by eldara-cruncher**
Fixes #611.

## The defect

The report is the inverse of its title: swarmcli *does* discover the switch — in some code paths and not others, and they disagree.

swarmcli held two independent notions of "the current Docker context":

- `docker.GetClient()` — a process-wide singleton built on first use and never rebuilt unless something calls `ResetClient()` (only the in-app switch did);
- `GetContextFromEnv`, `GetCurrentContext` and `GetDockerContext` — three near-identical resolvers, each shelling out to `docker context show` on **every call**.

`docker context use b` in another terminal moved the second and not the first. From then on:

| surface | addressed |
|---|---|
| stacks/services/nodes lists, log streaming | swarm **a** (cached client) |
| raw-CLI log fallback (`docker/service.go`, `views/logs/stream.go`) | swarm **b** |
| `stack deploy`, `stack rm` (`docker/stack_deploy.go`) | swarm **b** |
| `docker stack services` / `service inspect` / `network ls` (`stack_to_compose.go`) | **the config file**, always — no `--context` at all, so these ignored `DOCKER_CONTEXT` too |
| contexts view's "current" marker | swarm **b** |
| header "Context:" | frozen (see below) |

The reported logs failure is `docker --context b service logs <id-from-a>`. **Deploying a stack shown from `a` into `b`, with no warning, is the same defect and worse.**

## The fix

**One session pin.** `docker/session_context.go` resolves the context once — `DOCKER_CONTEXT`, else `docker context show` — and holds it. `buildClient` builds from it, the three resolvers return it, `ListContexts` recomputes `Current` against it instead of trusting `docker context ls`, and the three unflagged `stack_to_compose` commands now name it with `--context`. Only a switch made inside swarmcli moves it.

`--context` rather than exporting `DOCKER_CONTEXT`: `docker/cli@v28.5.1 cli/command/cli.go:461` resolves `--context` > `$DOCKER_HOST` (which forces `default`) > `$DOCKER_CONTEXT`, so the variable is not a reliable pin for a child process. The flag is.

**A prompt, not silence.** A deliberate switch should not be ignored either, so a tick compares the config file against the pin and offers to follow:

```
The Docker context changed outside swarmcli: 'a' → 'b'.

swarmcli is still using 'a'. Switch to 'b'?

y — switch to 'b' and reload the cluster
n — keep using 'a' (swarmcli will not ask again for 'b')
```

- Declining is remembered **per target**, so two terminals on two swarms means one question, not one every five seconds. A drift to a *third* context still asks, and returning to the pin clears the refusal.
- `DOCKER_CONTEXT` being set is never drift — it outranks the config file by design, and `ValidateContext` refuses to move off it anyway (#563/#565). Prompting there would offer a switch the switcher then refuses, at every startup of the DinD test flow.
- The prompt waits for the screen: no dialog over the startup overlay, an app error, the unlock dialog or the update notice — and the update notice now yields to it in return.
- Accepting does **not** run `docker context use` (the config file already says `b`), and a revert after a failed load writes `~/.docker/config.json` only when swarmcli was the one that wrote it. Undoing the operator's own shell switch would be reaching outside our own state.
- Both ways into a new context now go through one `enterContext` helper, so the client drop, the snapshot invalidation, the loading view and the revert bookkeeping cannot drift apart.

**An adjacent defect the prompt needed fixed.** `app/tick.go`'s `handleTick` never returned `tick()`, so the timer armed in `Init` fired once and stopped — untouched since `app` was extracted from `main` in #26. That is why the header's context name *and* its container/service counts froze seconds after launch. The drift check needs a live timer, and the header was always meant to have one. Note this means `LoadStatus` (two API calls against the cached client) now runs every 5 s where it effectively ran once.

## Tests

24 new unit tests, no daemon required. The regression test for this issue is `TestSessionContext_DoesNotFollowTheConfigFile` plus `TestResolvers_AllReportThePin`: resolve, move what `docker context show` would return, assert all three resolvers still agree on the first answer. Also covered: the env pin, `SetSessionContext`, a failed lookup not being cached, `ListContexts` marking the pin over `docker context ls`'s own flag, the three `--context` arg builders, the whole drift state machine (raise / decline / third target / return clears / waits for the screen / accepts), the prompt being a `y/n` confirmation that owns the keyboard, the revert not writing the config file, and `handleTick` re-arming.

`integration-tests/docker/get_context_test.go` gains a daemon-backed half — a real `docker context use` against the real config file must not move a resolved session — and both it and `swarmlock` now drop the pin where they change what resolution would return.

## Deviation from the plan

`InitSessionContext()` is called from `app.InitialModel()`, not `app.Init()` or `cli.Dispatch()`: `Init()` also runs for `swarmcli --version`, which has no business spawning a `docker` subprocess. The lazy path in `SessionContext()` covers the non-interactive CLI and library consumers, and a short-lived CLI process cannot drift anyway.

## Verification

`go build ./...`, `go vet ./...`, `go vet -tags integration ./...`, `go test ./...`, `golangci-lint run ./... --build-tags=integration` — all clean. Integration tests not run here (no DinD swarm in this environment).

---
Machine-generated by an AI agent (Claude Opus 5) and not human-reviewed. Claims carry a file, a line or a reproducible check so they can be verified cheaply.

### DIFF
--- a/app/contextdrift.go
+++ b/app/contextdrift.go
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package app
+
+import (
+	"fmt"
+
+	tea "github.com/charmbracelet/bubbletea"
+
+	"github.com/Eldara-Tech/swarmcli/v2/docker"
+)
+
+// The Docker context is pinned for the life of the session (docker.SessionContext),
+// so a `docker context use` run in another terminal moves nothing here. That is
+// the fix for #611 — but a switch the operator made deliberately should not be
+// silently ignored either, so the app watches for the divergence and offers it.
+//
+// Watching, not following: the prompt is the only thing that moves the pin, and
+// declining is remembered, so an operator who keeps two terminals on two swarms
+// is asked once per target rather than every five seconds.
+
+// contextDriftMsg reports the result of one drift check. Shell is the context
+// the config file now names, or "" when it agrees with the session pin.
+type contextDriftMsg struct{ Shell string }
+
+// Seams, so the drift logic is testable without a Docker CLI on PATH.
+var (
+	sessionContextFn    = docker.SessionContext
+	configFileContextFn = docker.ConfigFileContext
+	setSessionContextFn = docker.SetSessionContext
+	envPinsContextFn    = docker.EnvPinsContext
+)
+
+// checkContextDriftCmd compares the context Docker would use now against the
+// one this session is pinned to.
+//
+// A failed lookup reports "no drift" rather than an error: `docker context show`
+// can fail transiently, and a dialog raised over a subprocess hiccup would be
+// worse than a check that quietly runs again five seconds later.
+func checkContextDriftCmd() tea.Cmd {
+	return func() tea.Msg {
+		// DOCKER_CONTEXT outranks the config file, so while it is set the two
+		// naming different contexts is the arrangement the operator asked for,
+		// not a switch. Offering to follow it would also be offering something
+		// the context switcher refuses to do.
+		if envPinsContextFn() {
+			return contextDriftMsg{}
+		}
+		pinned, err := sessionContextFn()
+		if err != nil || pinned == "" {
+			return contextDriftMsg{}
+		}
+		shell, err := configFileContextFn()
+		if err != nil || shell == "" || shell == pinned {
+			return contextDriftMsg{}
+		}
+		return contextDriftMsg{Shell: shell}
+	}
+}
+
+// handleContextDrift decides what one drift check means for the UI.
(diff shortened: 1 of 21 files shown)


mosonyi:
LGTM!

Issue #601, opened 2026-08-20
Change color at services


mosonyi:
If I check the tasks of a service, I can find exited ones (a proper shutdown, without errors)

They should be yellow for instance. Use red only at failure.

<img width="1015" height="183" alt="Image" src="https://github.com/user-attachments/assets/c1d1f4fa-b1fd-42c1-9344-e4da2598a5d6" />

PR #615, opened 2026-08-27
fix(views): a task that ran to completion is not a warning

### LINKED ISSUE #613: Exit code 0 show red
Why do I see red when the container is already in completed without any error?

Image

eldara-cruncher:
An expanded service listed every finished job task in the "stopped" colour, so a migration that had done exactly what it was asked read like something to look at. On a job service — `restart_policy: none` or `on-failure`, which is what `isJobService` already calls a job — the entire task history is `complete`, so the tint applied to every row and stopped meaning anything, the way red did before #602.

### Why `complete` is different from `shutdown`

`complete` is the one swarm task state that proves a clean exit. The agent reaches it only by `transition(api.TaskStateCompleted, "finished")` after `ctlr.Wait(ctx)` returns nil (`swarmkit@v2.1.2 agent/exec/controller.go:322-327`), and the container executor's `Wait` returns nil only when `status.ExitCode() == 0` — anything else is wrapped in an `exitError` and becomes `fatal(err)` → `failed` with `task: non-zero exit (N)` (`docker@v28.5.2 daemon/cluster/executor/container/controller.go:321-339`).

#602 filed `complete` into `isStoppedState` as "over without having failed". That is true, and not the point: `shutdown` is swarm ending a replica that meant to keep running, `complete` is the task ending itself successfully. The rest of the tree already draws that line — `docker/convergence.go:114` counts a `complete` job task as *completed*, and both error detectors explicitly skip `DesiredState=shutdown && State=complete` rather than flag the service (`views/services/update.go:693`, `views/stacks/update.go:1317`).

### The change

`complete` returns the caller's own row style, and `isStoppedState` is left to the two states swarm ends. It needs an arm of its own rather than just leaving that helper, because the container's docker-ps state is `exited` here too and the `t.ContainerState == "exited"` clause would tint the row regardless. The arm sits *after* the failure case, so a `complete` task carrying an `Error`, `Health == "unhealthy"` or `ContainerState == "dead"` is still red.

Red stays for failed / rejected / orphaned / error / unhealthy. Yellow stays for shutdown / remove / starting / restarting, so #601 is unaffected.

Nothing is lost by not tinting it: the CURRENT STATE column already reads `complete 44 seconds ago` next to `running 18 minutes ago`.

### Tests

`views/taskutil/style_test.go` flips the `complete` case and adds two: a completed task whose container has already been pruned off the node, and a `complete` row carrying an error, which pins the arm order so a later reshuffle of the switch can't go unnoticed.

`views/services/columns_test.go` grows the two rows the issue was actually reported on — same state, differing only in whether the health decorator still has the container in its inventory. In the reporter's screenshot those two groups rendered in different colours; they must not.

Both new assertion sets were checked against the unfixed rule and fail there.

### Note on the reported screenshot

The colour in #613 is red rather than yellow because that build predates #602: sampling the image gives `#E74856` (colour 9) for the tinted rows and `#CCCCCC` (colour 7) for the untinted ones, which is the old `taskRowStyle(firstNonEmpty(Health, ContainerState))` rule keying on the container state — the new rule keys on the task state and would have tinted all five rows alike. #602 shipped in v2.0.0-rc2; this PR removes the yellow it left behind.

Closes #613

### DIFF
--- a/views/services/columns_test.go
+++ b/views/services/columns_test.go
@@ -159,8 +159,10 @@ func TestExpandedRow_ContainerStateFallback(t *testing.T) {
 }
 
 // The tint rule itself is tested in views/taskutil; what this asserts is that
-// the expanded rows go through it, and that a replica swarm stopped on purpose
-// no longer reads like the one that crashed next to it (issue #601).
+// the expanded rows go through it, that a replica swarm stopped on purpose no
+// longer reads like the one that crashed next to it (issue #601), and that a
+// task which ran to completion reads like an ordinary row whether or not its
+// container is still on the node to report "exited" (issue #613).
 func TestExpandedRow_TintsByTaskState(t *testing.T) {
 	trueColour(t)
 	m := testModel()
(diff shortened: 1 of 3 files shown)


mosonyi (approved):
lgtm

PR #633, opened 2026-09-02
fix(services): HEALTH stops guessing once the task is over

### LINKED ISSUE #616: services: HEALTH shows the container state on tasks that are already over, inconsistently
Split out of #613, where it is visible but is not what was reported.

The HEALTH column in an expanded service carries the healthcheck verdict — `healthy` / `unhealthy` / `starting`. When the image declares no `HEALTHCHECK` that field is empty, so the cell falls back to the container's live docker-ps state:

```go
// views/services/update.go:851
health: dashIfEmpty(TruncateRunes(firstNonEmpty(task.Health, task.ContainerState), 9)),
```

The comment above it says the fallback exists "so container errors surface even for images without a healthcheck", and for a **running** task it earns that: the task claims `running`, the cell says `restarting`, and an operator has learned something real.

For a **terminal** task it says nothing. The task is `complete`; of course the container exited. Worse, it is inconsistent between rows in the same state. In the screenshot on #613 all five tasks are `complete`, three read `exited` and two read `—`. The only difference between the two groups is whether the health decorator still has that container in its inventory — once the node has pruned the exited container, the lookup misses and the cell empties. So the column advertises a distinction between rows where none exists, and `exited` is the word that makes the first three look like the interesting ones.

**Suggested behaviour.** Once the swarm task state is terminal (`complete` / `shutdown` / `failed` / `rejected` / `orphaned`), stop falling back: show the healthcheck verdict if one was recorded, otherwise `—`. The column then means only "what the healthcheck says", which is what its header claims.

This does not weaken the row tint. `views/taskutil.TaskRowStyle` reads `ContainerState` directly, not the rendered cell, so `dead` still tints red and `restarting` still tints yellow whatever the column shows.

Scope is `views/services/update.go:851` plus `TestExpandedRow_ContainerStateFallback` and neighbours in `views/services/columns_test.go`.

---
Machine-generated by an AI agent (Claude Opus 5) and not human-reviewed. Claims carry a file, a line or a reproducible check so they can be verified cheaply.

eldara-cruncher:
Closes #616.

## The column was reporting inventory, not health

An expanded service's HEALTH cell shows the healthcheck verdict, falling back to
the container's live docker-ps state when the image declares no `HEALTHCHECK`.
For a **running** task that fallback earns its place — the task claims `running`,
the cell says `restarting`, and an operator has learned something real.

For a task that is **over** it says nothing: of course the container exited. And
it does not even say it consistently. In the screenshot on #613 all five tasks
are `complete`, three read `exited` and two read `—`. The only difference is
whether the node has pruned the container yet — `FanOutTaskOps.enrich` stamps
`Health` and `ContainerState` only on an inventory hit
([`swarmcli-be internal/ctrclient/fanout.go:330`](https://github.com/Eldara-Tech/swarmcli-be/blob/main/internal/ctrclient/fanout.go#L330)),
so a pruned container leaves both empty. The column advertised a distinction
between rows where there is none between tasks.

## What changed

`taskHealthCell` is now the single answer to "what has HEALTH got to say about
this task": the recorded verdict when there is one, else the container's live
state — and that fallback stops once `taskutil.IsTerminal(t.State)`.

It returns `Health` rather than `""` for a terminal task on purpose. `Health` is
a documented decorator extension point; today's agent derives it by parsing the
docker-ps status parenthetical
([`swarmcli-agent cmd/agent/containers.go:77`](https://github.com/Eldara-Tech/swarmcli-agent/blob/main/cmd/agent/containers.go#L77)),
which on an exited container holds the exit code (`Exited (0) 3 minutes ago`) and
never a health token — so in practice a terminal task renders `—`. A decorator
reading `ContainerInspect` would keep a real post-mortem verdict there, and this
cell should show it.

**The column-visibility gate is the other half of the fix**, and is the part #616
does not name. It tested the raw fields, so a service whose tasks are all over —
every job service, and the one on #613 — would have kept the column and filled it
with a dash on every row. Gate and cell now both go through `taskHealthCell`, so
they cannot drift:

| #613's service, 5 `complete` tasks, no healthcheck | HEALTH column |
| --- | --- |
| before | `exited  exited  exited  —  —` |
| cell fix alone | `—  —  —  —  —` |
| this PR | column not rendered |

`IsTerminal` joins `isFailedState` and `isStoppedState` in `taskutil`, composed
from them plus `complete`, which belongs to neither. It covers `remove` as well
as the five states #616 lists: `isStoppedState` already pairs it with `shutdown`,
and a task on its way out of the task list has an exited-or-gone container like
the rest.

## Not affected

The row tint. `taskutil.TaskRowStyle` reads `ContainerState` directly rather than
the rendered cell, so `dead` still tints red and `restarting` still tints yellow
whatever the column shows. `TestExpandedRow_TintsByTaskState` — which is #613's
own five-row fixture — is unchanged and still green.

Only the services view has a task HEALTH column; the stacks and tasks views
render task sub-rows without one.

## Verification

- `TestIsTerminal` — the six terminal states against seven that may still run.
- `TestExpandedRow_HealthDropsWhenEveryTaskIsOver` — the #613 shape: no HEALTH
  header, and `exited` nowhere in the render. Fails on `main` with the rendered
  row printed in the diff.
- `TestExpandedRow_HealthKeepsContainerStateWhileTaskCouldRun` — one `running`
  task whose container has died beside one `complete` task: the column returns,
  `exited` appears on the live row only. This is the case the fallback exists for.
- `TestExpandedRow_ContainerStateFallback` now pins `State: "running"`; it was
  passing by omission, since an empty state is not terminal either way.
- `go test ./...`, `go vet ./...`, `gofmt` and `scripts/check-spdx.sh` clean.
  `golangci-lint` could not run locally (the installed binary is built with
  go1.26, below the repo's 1.27 target since #631) — leaving that to CI.

---
Machine-generated by an AI agent (Claude Opus 5) and not human-reviewed. Claims
carry a file, a line or a reproducible check so they can be verified cheaply.

---

## Follow-up from review: the service row's ERROR (`mosonyi`, 2026-09-09)

> It shows the correct run of the last task. But why it reports error in general?

Right, and it is a second defect rather than a leftover of this one. In the
screenshot `zammad_init` exits 0 three minutes ago and the row above the tasks
still reads `task: non-zero exit (1)` — the attempt from a minute before it.

**Why.** A slot's failure was only ever cancelled by a **running** task newer
than it. A run-to-completion service never has one: it exits, is not restarted,
and sits at `0/1` for good. The `desired > 0 && running < desired` gate feeding
the error scan is therefore permanently open for it, the scan takes any failure
within 5 minutes of the service's newest task, and nothing checks that a later
task already succeeded. Both the services view's ERROR column and the stacks
badge did this, in near-identical copies, and the cutoff is anchored to a task
timestamp rather than the clock — so it never ages out either.

**Fix.** `taskutil.ActiveDeploymentErrorsByService` now counts a task that
reached `complete` as a sign of life alongside a running one. That is the whole
correction; the per-view scans were the same rule with a cruder supersession
test, so both views now call the shared helper:

| `zammad_init`, newest run `complete` 3m ago, `failed` 4m ago | service row |
| --- | --- |
| before | red, `task: non-zero exit (1)` |
| after | not failing, ERROR empty |
| newest run `failed` instead | red, that failure's message |

226 lines go with it. `svcDesired`/`svcRunning` existed only to gate the scan, so
the badge no longer counts replicas at all — which means it cannot recover #643's
false positive by construction, and the line #644 added there last night is gone
with the rest.

**One judgement call.** The two paths disagreed about the 5-minute window: the
per-view scan measured it from the service's newest task, the shared helper from
the cluster's. The per-service anchor wins, so a service that broke weeks ago and
has not been rescheduled since keeps saying so however busy its neighbours are.
`swarmcli-be`'s `bootstrap.ServiceHealth` reads the same helper and inherits both
changes — it builds and `go test ./bootstrap/...` passes against this branch.

Two smaller consequences, both improvements: the stacks badge walks
`snap.Services` rather than a map, so a stack with several failing services names
the same one on every refresh; and a failed task carrying no message now reads
`task failed` instead of tinting the row red behind an empty ERROR cell.

**Verification.** `TestActiveDeploymentErrors_CompletedRunSupersedesFailure` and
`TestActiveDeploymentErrors_WindowFollowsTheService` both fail on the previous
implementation; `TestServiceErrors_*` and `TestStackErrors_*` drive the
screenshot's shape through each view with a passing control where the newest run
is the failure. `go test ./...`, `go vet ./...`, `gofmt` and
`scripts/check-spdx.sh` clean.

This branch also carries a merge of `main` (it was 9 commits behind, and #644
touched the same block).

### DIFF
--- a/docker/task.go
+++ b/docker/task.go
@@ -48,7 +48,9 @@ type TaskEntry struct {
 	// decorator. Unlike State (the swarm task state) it reflects the
 	// container's `docker ps` state, which the remote Swarm API cannot report;
 	// the services view shows it as a fallback when Health is empty so container
-	// errors surface even for images without a healthcheck.
+	// errors surface even for images without a healthcheck — but only while the
+	// task could still be running, since a task that is over has an exited
+	// container by definition.
 	ContainerState string
 	// PullProgress summarizes the image pull the task's node is currently
 	// performing for it (e.g. "pulling · 3/12 layers · 412 MB"); "" when nothing
--- a/views/services/columns_test.go
+++ b/views/services/columns_test.go
@@ -143,12 +143,13 @@ func TestColumns_StatusShowsRolloutProgress(t *testing.T) {
 func TestExpandedRow_ContainerStateFallback(t *testing.T) {
 	m := testModel()
 	loadWithFilter(m, AllFilter, fakeEntries("web"))
-	// Expand the service and give it a replica that is erroring with no
-	// healthcheck (Health empty). The container's live state must still surface
-	// as the HEALTH fallback so failures show for images without a HEALTHCHECK.
+	// Expand the service and give it a replica that swarm still calls running
+	// while its container has died, with no healthcheck (Health empty). The
+	// container's live state must surface as the HEALTH fallback so failures
+	// show for images without a HEALTHCHECK.
 	m.expandedServices["id-web"] = true
 	m.serviceTasks["id-web"] = []docker.TaskEntry{{
-		Name: "web.1", NodeName: "node-1", DesiredState: "Running",
+		Name: "web.1", NodeName: "node-1", DesiredState: "Running", State: "running",
 		CurrentState: "running 8m", ContainerID: "c1", ContainerState: "exited",
 	}}
 	m.setRenderItem()
@@ -193,6 +194,52 @@ func TestExpandedRow_TintsByTaskState(t *testing.T) {
 	require.Equal(t, fgSeq(lipgloss.Color("7")), rowTint(t, out, "node-old"))
 }
 
+// Once a task is over, CURRENT STATE has already said so and the container's
+// "exited" adds nothing — worse, whether the cell can say it at all turns on
+// whether the node has pruned the container out of the health decorator's
+// inventory, which is a difference between rows and not between tasks. With no
+// healthcheck verdict left to show, the column goes away rather than filling
+// with dashes (issue #616).
+func TestExpandedRow_HealthDropsWhenEveryTaskIsOver(t *testing.T) {
+	m := testModel()
+	loadWithFilter(m, AllFilter, fakeEntries("web"))
+	m.expandedServices["id-web"] = true
+	m.serviceTasks["id-web"] = []docker.TaskEntry{
+		{Name: "web.1", NodeName: "node-done", DesiredState: "Shutdown", State: "complete",
+			CurrentState: "complete 44 seconds ago", ContainerState: "exited"},
+		{Name: "web.2", NodeName: "node-old", DesiredState: "Shutdown", State: "complete",
+			CurrentState: "complete 3 days ago"},
+	}
+	m.setRenderItem()
+
+	out := m.List.RenderItem(m.List.Filtered[0], false, 0)
+	require.NotContains(t, out, "HEALTH", "no task can report health, so the column earns no width")
+	require.NotContains(t, out, "exited", "a finished task's container state is not a healthcheck verdict")
+}
+
+// The fallback keeps earning its place while a task could still be running:
+// there the container's state contradicts the task's, and that is the container
+// error #616 was careful not to hide.
+func TestExpandedRow_HealthKeepsContainerStateWhileTaskCouldRun(t *testing.T) {
+	m := testModel()
+	loadWithFilter(m, AllFilter, fakeEntries("web"))
+	m.expandedServices["id-web"] = true
+	m.serviceTasks["id-web"] = []docker.TaskEntry{
+		{Name: "web.1", NodeName: "node-loop", DesiredState: "Running", State: "running",
+			CurrentState: "running 2 minutes ago", ContainerState: "exited"},
+		{Name: "web.2", NodeName: "node-done", DesiredState: "Shutdown", State: "complete",
+			CurrentState: "complete 44 seconds ago", ContainerState: "exited"},
+	}
+	m.setRenderItem()
+
+	out := m.List.RenderItem(m.List.Filtered[0], false, 0)
+	require.Contains(t, out, "HEALTH", "one task with something to report brings the column back")
+	require.Contains(t, rowFor(t, out, "node-loop"), "exited",
+		"a task swarm calls running whose container has died is the case the fallback exists for")
+	require.NotContains(t, rowFor(t, out, "node-done"), "exited",
+		"the finished task alongside it still says nothing")
+}
+
 // trueColour makes a test's assertions run against the tinted rows: the default
 // profile under `go test` is Ascii, where every style renders as plain text.
 func trueColour(t *testing.T) {
@@ -205,9 +252,15 @@ func trueColour(t *testing.T) {
 // rowTint returns the opening SGR sequence of the rendered row carrying marker.
 func rowTint(t *testing.T, out, marker string) string {
 	t.Helper()
+	return sgrPrefix.FindString(rowFor(t, out, marker))
+}
+
+// rowFor returns the rendered line carrying marker, styling included.
+func rowFor(t *testing.T, out, marker string) string {
+	t.Helper()
 	for _, line := range strings.Split(out, "\n") {
 		if strings.Contains(ansi.Strip(line), marker) {
-			return sgrPrefix.FindString(line)
+			return line
 		}
 	}
 	require.FailNowf(t, "no rendered row", "no row containing %q", marker)
--- a/views/services/update.go
+++ b/views/services/update.go
@@ -809,17 +809,17 @@ func (m *Model) setRenderItem() {
 			if len(tasks) > 0 {
 				// HEALTH and PORTS are per-container data the swarm API does not
 				// expose; they are populated by a TaskOps decorator and stay
-				// empty otherwise, so only show those columns when present. The
-				// HEALTH cell shows the healthcheck token when present, else the
-				// container's live state (running / restarting / exited), so the
-				// column appears whenever the decorator is active.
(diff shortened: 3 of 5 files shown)


mosonyi (changes requested):
Almost good. It shows the correct run of the last task.

But why it reports error in geenral?

Image

eldara-cruncher:
Good catch — that is a second defect, not a leftover of this one, and it is now fixed on this branch in 2ba1aee.

The row was never reading the last task. A slot's failure was only ever cancelled by a **running** task newer than it, and `zammad_init` never has one: it exits 0, is not restarted, and sits at `0/1` for good. So the "under-replicated" gate feeding the error scan stays permanently open, the scan picks any failure within 5 minutes of the service's newest task, and nothing checks that a later task already succeeded. The `failed 4 minutes ago` attempt is one minute inside that window. The cutoff is measured from a task timestamp rather than the clock, so it would have stayed there until the service next ran at all.

`taskutil.ActiveDeploymentErrorsByService` now counts a task that reached `complete` as a sign of life alongside a running one. The services view and the stacks badge carried near-identical copies of the same scan with a cruder version of that test, so both now call the shared helper — 226 lines lighter, and the two can no longer disagree about whether a service is failing.

`TestServiceErrors_CompletedRunClearsTheEarlierFailure` and `TestStackErrors_CompletedRunClearsTheEarlierFailure` drive your screenshot's shape through each view and fail on the previous implementation; each has a control asserting that when the newest run *is* the failure, the row still says so. Details and the one judgement call it forced (which task the 5-minute window is measured from) are in the PR description.

eldara-cruncher:
One more thing from the same screenshot, and this one is a question for you rather than a fix.

`zammad_init` reads `0/1`, and it always will: a one-shot service exits and is never restarted, so it is permanently "short" a replica it is not supposed to have. `docker service ls` prints `0/1` for it too, and `ReplicasOnNode` deliberately mirrors that ([`docker/service.go:357`](https://github.com/Eldara-Tech/swarmcli/blob/main/docker/service.go#L357)).

But we do already recognise the shape. `isJobService` ([`docker/convergence.go:185`](https://github.com/Eldara-Tech/swarmcli/blob/main/docker/convergence.go#L185)) spots exactly this — a replicated service with `restart_policy: none`/`on-failure`, which is the only way to express a one-shot step in a compose stack — and `--wait` already treats a completed job as converged rather than as under-replicated.

So which do you want in the services view: keep matching `docker service ls` verbatim, show completions the way swarm's native job modes do (`0/1 (5/5 completed)`), or leave REPLICAS alone and let STATUS say `complete`?

Nothing is broken either way, so "leave it" is a fine answer. I'll open an issue for whichever you pick.

mosonyi:
> One more thing from the same screenshot, and this one is a question for you rather than a fix.
> 
> `zammad_init` reads `0/1`, and it always will: a one-shot service exits and is never restarted, so it is permanently "short" a replica it is not supposed to have. `docker service ls` prints `0/1` for it too, and `ReplicasOnNode` deliberately mirrors that ([`docker/service.go:357`](https://github.com/Eldara-Tech/swarmcli/blob/main/docker/service.go#L357)).
> 
> But we do already recognise the shape. `isJobService` ([`docker/convergence.go:185`](https://github.com/Eldara-Tech/swarmcli/blob/main/docker/convergence.go#L185)) spots exactly this — a replicated service with `restart_policy: none`/`on-failure`, which is the only way to express a one-shot step in a compose stack — and `--wait` already treats a completed job as converged rather than as under-replicated.
> 
> So which do you want in the services view: keep matching `docker service ls` verbatim, show completions the way swarm's native job modes do (`0/1 (5/5 completed)`), or leave REPLICAS alone and let STATUS say `complete`?
> 
> Nothing is broken either way, so "leave it" is a fine answer. I'll open an issue for whichever you pick.

leave it

mosonyi:
Still not ok

image

eldara-cruncher:
Found it, and it is a different fault from the last one — fixed in b2f4b6a.

The rule I added was right (a later success clears an earlier failure), but it was reading the wrong clock. It compared `Status.Timestamp`, which says *when a task last changed state*, not *which attempt came later*. Those two part company as soon as a finished task's status is touched again — and every task in your screenshot carries `DesiredState: shutdown`, which swarm stamps onto the leftovers when the service is updated. A failure restamped at that moment outranks the run that actually succeeded after it.

I reproduced it both ways. With the timestamps in attempt order the previous commit already suppressed your `migrate` shape; only the restamped ordering brings `task: non-zero exit (1)` back. So the fix is to order attempts by `CreatedAt`, which is the clock that answers "which attempt came later" — and, not coincidentally, the one `GetTasksForService` sorts the expanded rows by. The service row and the rows beneath it now share one ordering, so the row cannot contradict the list, which is what you have reported twice.

`LatestTasksByServiceKey` picked its per-slot winner off the same wrong clock, so it moved too.

`TestServiceErrors_RestampedFailureDoesNotOutrankANewerRun` drives your exact shape — three failed attempts, then two completed runs, with the last failure restamped after both — through the services view, and fails on the previous commit.

One favour, so I can confirm the mechanism rather than infer it: could you paste the output of

```
docker service ps --no-trunc eldara-swarmcli_migrate
docker service ps eldara-swarmcli_migrate --format '{{.ID}} {{.Image}} {{.CurrentState}} {{.Error}}'
```

If a failed task's timestamp really does sit after the completed runs', that confirms the restamping; if it does not, the row was being drawn by a build from before 2ba1aee and I would rather know that than guess.

mosonyi (approved):
LGTM

PR #634, opened 2026-09-02
fix(secrets): the create dialog accepted names Swarm then refused

eldara-cruncher:
Refs Eldara-Tech/swarmcli-be#353 — the CE half of that investigation.

### The gap

`validateSecretName` (`views/secrets/model.go`) checked emptiness, whitespace and `/\:*?"<>|`. The daemon applies something different — swarmkit's `validateConfigOrSecretAnnotations`:

- at most **64** characters
- `^[a-zA-Z0-9]+(?:[a-zA-Z0-9-_.]*[a-zA-Z0-9])?$` — letters, digits, `-`, `_` and `.`, starting and ending alphanumeric

The create dialog's name input allows 100 characters, so a 65-character name was accepted by the dialog, sent, and refused by the daemon — the operator learning it was too long from a raw daemon error after the dialog had already closed. A leading dash or a trailing dot behaved the same way.

`validateConfigName` was the same function with the same gap, and configs go through the same daemon-side validator.

### The change

`docker.ValidateSwarmObjectName(kind, name)` holds swarmkit's rule once; both dialogs ask it, and both keep their own thin wrapper so the call sites and their tests are untouched. The messages name the object, so the configs dialog does not talk about secrets:

```
secret name must be 64 characters or fewer (this one is 65)
secret name may contain only letters, digits, '-', '_' and '.', and must start and end with a letter or a digit
```

The input's `CharLimit` is deliberately left at 100: capping it at 64 would silently swallow the tail of a pasted name, where the message says what is wrong.

### Tests

`docker/names_test.go` covers the accepted set (including a dotted name and one exactly 64 long) and each rejection with the reason it gives. `views/secrets` and `views/configs` each gain a length case on their validator, and the secrets create dialog gains one asserting an over-long name is reported in the dialog rather than by the daemon.

### Not in this PR

The bug that started the investigation is in swarmcli-be: revealing a secret builds a service named `swarmcli-reveal-<secret>-<unix>`, and Swarm holds a *service* to 63 characters and to a DNS label — a stricter rule than the one its secret name passed. Fixed separately in `swarmcli-be`.

---
Machine-generated by an AI agent (Claude Opus 5) via the `eldara-cruncher` bot account, and not human-reviewed. Claims carry a file, a line or a reproducible check so they can be verified cheaply.

### DIFF
--- a/docker/names.go
+++ b/docker/names.go
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright © 2026 Eldara Tech
+
+package docker
+
+import (
+	"fmt"
+	"regexp"
+)
+
+// MaxSwarmObjectNameLen is the length swarmkit's
+// validateConfigOrSecretAnnotations allows a secret or config name. Services
+// and networks are held to a stricter 63 by a different validator, so this is
+// not a limit for swarm objects in general.
+const MaxSwarmObjectNameLen = 64
+
+// swarmObjectNameRe is swarmkit's own expression for a secret or config name.
+var swarmObjectNameRe = regexp.MustCompile(`^[a-zA-Z0-9]+(?:[a-zA-Z0-9-_.]*[a-zA-Z0-9])?$`)
+
+// ValidateSwarmObjectName reports why the daemon would refuse this name for a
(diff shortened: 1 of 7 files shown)


mosonyi (approved):
LGTM

PR #644, opened 2026-09-07
fix(docker): count only constraint-eligible nodes for a global service

### LINKED ISSUE #643: Max Replicas at swarmcli view is wrong.
### SwarmCLI version

2.0.0

### OS / platform

rocky linux

### Steps to reproduce

1. create a swarm cluster with 3 managers and 3  workers. 
2. deploy a servivce  with mode:global and placement -node.role == manager 
 ```
deploy:
      mode: global
      placement:
        constraints:
          - node.role == manager
```

the problem is always with the global placement.

### Expected behaviour

the max Replicas should be the number of node that satisfying the  specific condition (manager. labeled, whatever...)

### Actual behaviour

the max replicas is always the number of nodes.

eldara-cruncher:
Fixes #643.

A global service's target count was the whole active node count, ignoring `TaskTemplate.Placement.Constraints`. A service pinned to `node.role == manager` on a 3+3 swarm read `3/6` and never looked converged — in the REPLICAS column, in the services and stacks error badges, and in `--wait`.

Swarm's global orchestrator filters twice: drop drained and down nodes, then `constraint.NodeMatches`. SwarmCLI implemented the first half only.

### Choices

**Ported swarmkit's constraint matcher instead of depending on it.** Its `manager/constraint` package matches against swarmkit's protobuf node type, not the API's `swarm.Node`, and the module drags grpc and raft behind it. `docker/placement.go` keeps the semantics exactly, including the parts that look like quirks: case-insensitive comparison, an absent label matching the empty string, and an unrecognised key matching no node. All keys are covered — `node.id`, `node.hostname`, `node.ip` (address or CIDR), `node.role`, `node.platform.os`, `node.platform.arch`, `node.labels.*`, `engine.labels.*` — so counts agree with `docker service ls`.

**Computed locally rather than reading `ServiceStatus.DesiredTasks` from the daemon.** That field counts tasks whose desired state is not Shutdown, so a downed node keeps counting for ~24h — which would undo #480 and hang `--wait` again.

**Replicated services are untouched.** A declared replica count is the target wherever the replicas can land. Filtering it would report an unschedulable service as converged, the opposite of what `--wait` is for.

**Widened the node predicate to swarm's own** — drained and down, nothing else, where SwarmCLI additionally demanded Ready and Active. The two halves of the ratio had come apart: a paused node's running task was counted in the numerator while the node was dropped from the denominator, so a healthy service read `3/2`.

**Deduplicated the stacks view.** It carried its own copy of the mode switch using `len(snap.Nodes)`, so it had this bug and #480's as well; it now calls the shared helper.

### Follow-up, not in this PR

`swarmcli-be/bootstrap/health.go` has the same duplicated switch and needs the same swap.

https://claude.ai/code/session_01EzRRCjxpsL5aCpomrcTCsA

### DIFF
--- a/docker/convergence.go
+++ b/docker/convergence.go
@@ -52,17 +52,33 @@ type ServiceConvergence struct {
 	NewestTaskAge time.Duration
 }
 
-// activeNodeIDs returns the nodes that can currently run tasks. A task pinned to
-// a node that is down never converges, so counting it in the denominator would
-// make a release hang until timeout for a reason no redeploy can fix.
-func activeNodeIDs(snap *SwarmSnapshot) map[string]struct{} {
-	active := make(map[string]struct{}, len(snap.Nodes))
+// schedulableNodes returns the nodes that can currently run tasks. A task pinned
+// to a node that is down never converges, so counting it in the denominator
+// would make a release hang until timeout for a reason no redeploy can fix.
+//
+// The predicate is swarmkit's own (orchestrator/global.updateNode): drained and
+// down, nothing else. A stricter one that also demanded Ready and Active split
+// the two halves of the ratio apart — a paused node's running task was counted
+// in the numerator while the node itself was dropped from the denominator, so a
+// healthy global service read 3/2.
+func schedulableNodes(snap *SwarmSnapshot) []swarm.Node {
+	out := make([]swarm.Node, 0, len(snap.Nodes))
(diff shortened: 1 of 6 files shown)


clangenb (approved):

mosonyi (approved):
Image

Looks good to me!

PR #646, opened 2026-09-10
Update README.md

aigents999:


README.md
```diff
@@ -1,317 +1,145 @@
...
 
----
-
-## Visual Feast
+<p align="center">
+  <a href="assets/swarmcli.gif"><img src="assets/swarmcli-screenshot.png" alt="SwarmCLI showing a stack's services, their tasks, and the node each runs on" width="800"></a>
+  <br /><sub>Click for the animated tour.</sub>
+</p>
 
-![alt text](assets/swarmcli.gif)
```
mosonyi, line 35: I liked better the visual gif on the main page.

README.md
```diff
@@ -1,317 +1,145 @@
...
-```
+## First run
 
-Then run:
+SwarmCLI needs a Docker context that reaches a swarm manager. Locally, that is the socket; for a remote swarm, a context:
 
-```
-go run .
+```bash
+docker context create prod --docker "host=ssh://ops@manager1.example.com"
```
mosonyi, line 64: Remove this, our goal is not to reach the context through ssh but the proxy with atomated managed context. Even though it is possible with ssh

mosonyi (changes requested):

PR #649, opened 2026-09-10
fix(charts): judge a one-shot by its task, not by the pause it cannot avoid

eldara-cruncher:
**Problem**
A one-shot's task ends inside the monitor window by design and swarmkit counts that as an update failure, so the default `failure_action` leaves every re-run job `paused` — which `Convergence()` read as `update paused after a task failure; swarm will not continue without intervention`.

**Fix**
A `Job` whose `UpdateState` is `paused` is judged by its own task, and the task counting drops the history that let a failed run inherit its predecessors' successes.

Reproduced on a real single-node swarm (docker 29.2.1): a stack with a `restart_policy: condition: none` service that exits 0, redeployed once, leaves

```
{"State":"paused","Message":"update paused due to failure or early termination of task u6qkbnvhsf18stifeee5mh50h"}
```

`UpdateStatus` is `null` after the *first* deploy, which is why this only appears from the second on. Upstream is explicit that there is no exemption: swarmkit's `updater.go` `failureTriggersAction` fires on any task that leaves `RUNNING` inside `UpdateConfig.Monitor`, and `TaskStateCompleted` is greater than `RUNNING`. So a migration that exited 0 and one that exited 3 leave the service in the identical state, and `charts/release.go` reported both as wedged.

The second defect is worse and was found while reproducing the first. `StackConvergence` counted `Complete` tasks from an unfiltered `TaskList`, which retains terminal tasks up to `--task-history-limit`. With three old `Complete` tasks and a newest one `Failed "task: non-zero exit (3)"`, `Completed` read 3 against `Desired` 1, so `Rollup` returned **converged** — a failed migration passing `--wait`. `TestStackConvergenceJudgesOnlyTheLatestJobRun` fails with `expected: 0, actual: 3` without the fix. It is also what the `5/1` and `11/8 tasks running` details in the `:charts` view were counting.

- `docker/service.go` — `newestTaskPerSlot` extracted out of `countUpToDateTasks`. The newest-per-slot rule and its rationale were already there for the services view's `UpToDate` column; `StackConvergence` asked the same question of the same task list and answered it differently.
- `docker/convergence.go` — the counting loop iterates that generation instead of raw `snap.Tasks`. This subsumes the `NodeID == ""` guard, since an unassigned task has no slot to be the newest of.
- `charts/release.go` — only `paused` is neutralised, and only for a `Job`. The rollback states still mean the spec actually moved back, so they stay wedged. A job that did *not* complete now says `the one-shot task did not complete (0/1); swarm paused the rollout` rather than sharing a sentence with the success case.

Both directions checked against the live swarm before teardown:

```
repro_migrate  job=true update="paused" running=0 completed=1 desired=1 -> converged
repro_migrate  job=true update="paused" running=0 completed=0 desired=1 -> wedged the one-shot task did not complete (0/1); swarm paused the rollout
```

`charts/backend_docker_test.go`'s `runningTask` gained a slot argument: it built two replicas of one service sharing slot 0, which real swarm never does — that is two generations of one replica, and the fixture only passed before because the count ignored generations.

`go vet` and the full suite pass. No integration coverage added, and the suite could carry it — `charts_wait_test.go` already deploys compose fixtures against the testenv swarm and asserts on `--wait`. What no fixture does is deploy a one-shot *twice*, which is what it takes to reach `paused` at all. Offered as a follow-up rather than folded in here.

</details>

https://claude.ai/code/session_01G1AbPB7Ao3b3SStJZKeM6h


mosonyi:
LGTM

PR #650, opened 2026-09-10
fix(charts): a one-shot swarm will not retry has failed, not stalled

eldara-cruncher:
**Problem**
#649 fixed two convergence defects no test could reach — the pause only appears on a redeploy, the false `converged` only once task history exists — and writing that coverage surfaced a third: a one-shot whose task swarm will not retry was reported as `0/1 tasks running` until the caller's timeout, never naming why.

**Fix**
Two integration tests against the DinD swarm, plus: a job under `restart_policy: none` whose task ended without completing is now wedged, carrying swarm's own reason.

**The third defect, and how it turned up.** The first version of these tests passed in isolation and failed in CI. Reproduced locally by running the whole package rather than just the two tests, and watched live: the task was `Rejected — "invalid pool request: Pool overlaps with other one on this address space"`, and it sat there. Every charts test installs and uninstalls a stack, so uninstall frees a /24 and the next install is handed it back before the node has torn the old one down. Every other service in the suite rides this out because the default restart policy replaces the task. A one-shot cannot: `condition: none` is the point of it, so the slot keeps a dead task forever.

swarmcli then reported that release as *progressing*. `Convergence()` now returns wedged for it, with `Status.Err` verbatim:

```
the one-shot task did not complete and swarm will not retry it: task: non-zero exit (3)
```

Restricted to `condition: none` deliberately. `on-failure` is a job to `isJobService` as well, but swarm replaces its failed tasks, so a failure there is not yet terminal — `TestStackConvergenceDoesNotBuryARetryableFailure` holds that line.

**The tests.** Both were run against the real 3-node testenv, and both fail against `860deae` (the commit before #649) — each for its own defect:

- `TestWaitAcceptsAOneShotSwarmPausedForFinishing` — `Received unexpected error … the second run also exited 0; --wait must not report the release wedged`
- `TestWaitRefusesAOneShotWhoseLatestRunFailed` — `Should be zero, but was 1 … the newest task failed; its predecessors are history, not progress`

The full package now passes end to end (`ok … 437.881s`), including the position where it failed twice before.

`monitor: 12s` is declared rather than left to swarm's 5s default because the window is the mechanism: swarmkit only counts a task that leaves RUNNING *inside* `UpdateConfig.Monitor` as an update failure, so on a slow runner a container finishing outside it would never pause and the test would prove nothing. `awaitRelease` is what stops that being a silent pass — it `t.Fatalf`s with the last state it saw. The second test names the exit status in its assertion for the same reason: a refused task also leaves the release wedged with `Completed 0`, and "not converged" alone would have tested nothing about task history.

The first test uses `on-failure` — still a job by `isJobService`, so identical coverage — so a refused task is retried rather than stranding the run.

**Not fixed here.** `--wait` on an *upgrade* can return before swarm has started the update. `deployAndRecord` (`charts/release.go`) runs `DeployStack` → `RefreshSnapshot` → `waitReady`, and `docker stack deploy` returns as soon as the manager accepts the spec, so the first poll can still see the previous generation: running, at parity, past its window. An earlier draft of the second test routed its assertion through `--wait` and was flaky for exactly this reason. It is the `#473` family rather than anything here, and it changes the hot path of every `--wait`, so it wants its own change.

</details>

https://claude.ai/code/session_01G1AbPB7Ao3b3SStJZKeM6h


eldara-cruncher:
Correction to the "Not fixed here" note above: **that defect does not reproduce, and the note is wrong.**

The claim was that `--wait` on an upgrade can return before swarm has started the update, because `deployAndRecord` runs `DeployStack` → `RefreshSnapshot` → `waitReady` and the first poll can still see the previous generation at parity.

Measured on a single-node swarm, sampling `charts.Rollup` every 100 ms across three upgrades of a converged one-shot to a revision that exits 3:

```
t+  0.07s  converged    update="completed" run=0 done=1 want=1
t+  3.77s  progressing  update="updating"  run=0 done=0 want=1   <- rolling update in progress
t+  4.96s  wedged       ... task: non-zero exit (3)
```

The sampler starts one second before the deploy is issued, so the only `converged` reading is the pre-deploy state. The window I described is negative rather than narrow: `DeployStack` shells out to `docker stack deploy` (`docker/stack_deploy.go:154`), which takes ~3 s, and swarm stamps `UpdateStatus=updating` before it returns — run 3 flipped at t+1.72s. `Convergence()`'s existing `case "updating"` arm therefore already holds the wait, and no `converged` reading occurs after the spec lands in any of the three runs.

The one observation behind the original claim — an `Upgrade(Wait: true)` returning nil on a revision whose task exited 3 — stays unexplained. It is not this mechanism, and I could not reproduce it. Reopening it would need that nil return caught again with the service state logged at the poll that produced it, not another reading of the code.

Unrelated but visible in the same trace: `dead=true` at t+4.96s is this PR working — the exit status is named within five seconds instead of `0/1 tasks running` until the caller's timeout.

Issue #659, opened 2026-09-15
config: a ~/.config/swarmcli/config.yaml that the UI can also write


eldara-cruncher:
**Problem**
Preferences exist only as environment variables (`SWARMCLI_TELEMETRY`, and `SWARMCLI_MOUSE` from #648). A choice made in the TUI, such as `:mouse`, is lost on the next start, and themes (#279) have nowhere to live.

**Fix**
Add one hand-editable `config.yaml` that swarmcli reads at startup. UI toggles write back to it one key at a time, without disturbing anything a person wrote.

<details><summary>Design</summary>

Decided:

- **Location.** `$XDG_CONFIG_HOME/swarmcli/config.yaml`, defaulting to `~/.config/swarmcli/config.yaml` on every OS, macOS included. That is next to the files already in that directory, and where gh and git look. Not `os.UserConfigDir`, which is `~/Library/Application Support` on darwin.
- **Preferences only.** Machine state (`install.json`, `update-notice.json`) stays in its own files; `telemetry/install.go` explains why.
- **Precedence: environment variable, then file, then default.** If a toggle saves a value that an environment variable overrides, the stack-bar line says so.
- **Toggles save immediately.** `:mouse` does, and later e.g. `:telemetry off`; the stack-bar line confirms the save.
- **Writes edit the yaml.v3 node tree**, as gh's `config set` does. Comments, key order and unknown keys survive; blank lines and the changed line's own inline comment do not. Never marshal a Go struct over the file: that is what k9s did, and it is behind derailed/k9s#261, #760, #2346 and #2421.
- **Re-read before every write.** Apply the single change to what is on disk now, so a hand edit made while swarmcli runs is kept. Write a temp file in the same directory and rename it over the original, keeping the file mode and following a symlink (dotfile managers).
- **Never write a file that does not parse.** swarmcli runs on defaults and says so once on the stack bar, and the toggle reports that it could not save.
- **Unknown keys are kept and logged, never fatal.** k9s v0.31 refused to start over a renamed key (derailed/k9s#2442). A bad value falls back to that key's default, except that `telemetry` keeps its rule that an unrecognised value leaves reporting on.
- **First write creates a commented template** listing every key.

Initial keys: `mouse` (bool) and `telemetry` (`on` | `off` | `none`). Themes (#279) add their own.

Open:

[...]

PR #660, opened 2026-09-15
feat(tui): mouse support: wheel, click to select, double click to open

### LINKED ISSUE #648: Investigate mouse click options
Currently we can navigate with keys.
Please investigate if we could do the same with mouse like in lazy docker.

eldara-cruncher:
**Problem**
SwarmCLI can only be driven from the keyboard; #648 asks for lazydocker-style mouse navigation.

**Fix**
The wheel moves the selection, a click selects a row or goes back to a breadcrumb, a double click does what Enter does and a right click does what Esc does; `SWARMCLI_MOUSE=off` or `:mouse` turns it off.

**How it fits together**

- **`app/mouse.go` handles every `tea.MouseMsg` itself.** No view ever receives a mouse event. Mouse input is ignored while a dialog, the startup overlay, the `:` bar, the `/` bar in edit mode, or a view's `CapturesInput` is holding the keyboard. Otherwise:
  - The wheel becomes up/down keypresses for the current view.
  - A left click becomes `view.RowClicker.ClickRow(line)`. The line is worked out from the same layout that `View` draws.
  - A second click on the same line within 500 ms becomes Enter.
  - A right click becomes Esc and goes through the app's Esc chain, so a passive `/` bar, a filter or an expanded row is closed before anything goes back.
  - A left click on the stack bar goes back to the clicked breadcrumb in one step (`returnTo`). The views it jumps over are not entered again, `…` goes to the nearest view it hides, and a passive `/` bar is closed first, as Esc would.
  - Only cell motion is requested, not all-motion.
- **New `view.RowClicker` seam plus `FilterableList.SelectLine`.** The seam is additive. Nodes, services, stacks, secrets and configs (plus their UsedBy lists), networks (plus UsedBy), volumes, charts and contexts implement it. Each repeats exactly what its own arrow-key path does after the cursor moves. A click anywhere inside an expanded stack, service or release selects the parent row.
- **`breadcrumbs()` gives each piece of the bar its click target.** `RenderBreadcrumbs` joins them, so the bar that is drawn and the bar that is hit-tested are one layout. Its output is byte-identical to before across 262,143 inputs.
- **`view.ExecProcess` / `view.Exec`.** Bubble Tea v1 does not turn mouse reporting back on after handing the terminal to another process (charmbracelet/bubbletea#1424). The three `$EDITOR` handoffs now go through these wrappers.
- **`app.ProgramOptions()`**, so an embedding build's `main` starts the same terminal.
- **`:mouse`** toggles mouse support for the session and confirms it on the stack bar. The first-run telemetry line still takes priority there.

**Checks**

- **Click geometry.** `TestClickLandsOnTheRenderedRow` finds each row where `View()` actually draws it and clicks there. It covers normal and fullscreen layouts, headers of 0, 1 and 2 lines, and the search bar open. Deliberately breaking the header offset, the double-click boundary or the dialog guard each makes the app tests fail.
- **Breadcrumbs and right click.** The tests click where `View()` draws the bar, with and without the search bar open. Eleven deliberate breakages each fail a test: the input-bar offset, the fullscreen guard, the width fit, entering jumped-over views, a clickable current crumb, the notice-only bar, the `…` target, the search bar close, the double-click reset, and right click skipping the Esc chain or the input guard.
- **Real binary in a pty**, against a one-node colima swarm with three probe configs (removed afterwards). 11/11 checks passed, twice:
  - startup sends `?1002h` and `?1006h`;
  - a click changes `Config k of 3`, and wheel up and down move it;
  - a double click opens inspect;
  - `:mouse` sends `?1002l`, and a second `:mouse` sends `?1002h`;
  - after an `EDITOR=true` handover, `?1002h` is sent again.
  - on a second run (6/6), a click on the ` configs ` breadcrumb in inspect returns to configs, and a right click there goes back to stacks. A control build without each fails exactly that check.
- **Negative control.** The same run with the restore removed fails exactly that last check.
- **Static checks.** `go test ./...`, `go vet ./...` (also with `-tags integration`), gofmt and `scripts/check-spdx.sh` are clean, and `golangci-lint run ./... --build-tags=integration` reports 0 issues.

**Trade-off.** While swarmcli has the mouse, native text selection needs Shift (Option in iTerm2). This is documented in the README and `docs/configuration.md`. Remembering the choice across restarts is #659.

**Not in this PR:** sorting by clicking column headers, and clickable help-bar entries. These were scoped out on #648. Right click and clickable breadcrumbs were added after review.

</details>

Closes #648

---

Machine-generated by an AI agent (Claude Opus 5) via the `eldara-cruncher` account,
and not human-reviewed. Claims carry a file, a line or a reproducible check so they
can be verified cheaply.

https://claude.ai/code/session_01FWetXJ6YoFuyhMEKYnVjnz


mosonyi (changes requested):
I like it, but please add:
Right click: ESC
Add the breadcumb as well.

eldara-cruncher:
@mosonyi both added in 0c68b0d.

A **right click** is Esc. It goes through the app's Esc chain, so it closes a passive `/` bar or a filter before it goes back. It follows the same guard as the wheel, so it does not close dialogs or the `:` and `/` bars while you type.

A **click on a breadcrumb** goes back to that view in one step. `…` goes to the nearest view it hides. With the mouse off (`:mouse`), neither happens.

---

Machine-generated by an AI agent (Claude Opus 5) via the `eldara-cruncher` account,
and not human-reviewed. Claims carry a file, a line or a reproducible check so they
can be verified cheaply.

mosonyi (approved):
LGTM

